PR #2, opened 2018-12-06
Renaming everything with an 'rcl_' prefix instead of an 'rc_' prefix …

nburek:
Renaming everything with an 'rcl_' prefix instead of an 'rc_' prefix to match the change in 'rcl'

Connects to https://github.com/ros2/rcl/pull/327


wjwwood:
Oh these are `COLCON_IGNORE`'d. So no CI should be needed.

wjwwood:
Actually wrong, just the log4cxx is ignored, here's Linux CI for the noop one:

[Build Status](https://ci.ros2.org/job/ci_linux/5819/)

PR #3, opened 2018-12-21
Enabling the build for rcl_logging_log4cxx

nburek:
This review includes all the fixes needed to build and run the rcl_logging_log4cxx package on Linux, Mac, and Windows 10. In order for this to work in the build system we will need use brew to install log4cxx on Mac and use the choco package in the related PR in order to install log4cxx on Windows. 

There is a known issue with the Windows build that causes log4cxx to crash when compiled as a Release build, but works if it is compiled in Debug mode. As a temporary work around we force the Windows build of rcl_logging_log4cxx to use Debug mode. 

Requires ros2/choco-packages#7

PR #4, opened 2019-03-08
[Crystal] Enabling the build for rcl_logging_log4cxx 

nuclearsandwich:
This is a backport of #3 to the Crystal release branch.

### DIFF
--- a/rcl_logging_log4cxx/CMakeLists.txt
+++ b/rcl_logging_log4cxx/CMakeLists.txt
@@ -12,15 +12,20 @@ if(NOT CMAKE_CXX_STANDARD)
 endif()
 
 find_package(ament_cmake_ros REQUIRED)
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+find_package(Log4cxx REQUIRED)
 
-set(dependencies
-    log4cxx
-)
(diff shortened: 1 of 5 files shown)


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6346/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2757/)
* macOS (on Crystal infra) [Build Status](https://ci.ros2.org/job/test_ci_osx/138/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6199/)

nuclearsandwich:
@nburek I'm not sure if there's a precise way to test that the rcl_logging_log4cxx is being loaded and used correctly. The CI above makes sure that it gets built and then tests the rcl package.

nburek:
@nuclearsandwich There is a variable set in the cmake files for which logger to load. Right now it is defaulted to rcl_logging_noop. So you would either need to change the default to use rcl_logging_log4cxx or set the environment variable to override the default when building rcl. 

After that you could test that rcl_logging_log4cxx is being loaded by running a demo application and looking for the log file in the default location (/var/log/ros for Linux). That is all very manual at the moment, but we could automate that test in the future by having a unit test that provides a different configuration to log4cxx to log to a temp file for the unit test to check after.

PR #9, opened 2019-05-02
Changing the default location for log files to be a local directory instead of /var/log/ros on linux due to permission issues.

nburek:
During the review of this PR: https://github.com/ros2/rcl/pull/425 we found that there were permission issues with writing to /var/log. The quick fix for this is to use the same behavior that we have on Windows, which is to log to a local directory instead.

### DIFF
--- a/rcl_logging_log4cxx/CMakeLists.txt
+++ b/rcl_logging_log4cxx/CMakeLists.txt
@@ -22,7 +22,7 @@ include_directories(${ament_INCLUDE_DIRS} include)
 if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 else()
-  # Disable ddl-interface warnings for STL from log4cxx code, because we are already 
+  # Disable ddl-interface warnings for STL from log4cxx code, because we are already
   # forced to use the same compiler in Windows, as we are getting the binaries
   # for log4cxx from Chocolatey
   # https://web.archive.org/web/20130317015847/http://connect.microsoft.com/VisualStudio/feedback/details/696593/vc-10-vs-2010-basic-string-exports
@@ -32,7 +32,6 @@ endif()
 set(${PROJECT_NAME}_sources
     src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
 )
-#set_source_files_properties(${${PROJECT_NAME}_sources} PROPERTIES language "CXX")
 
 if(NOT WIN32)
   add_library(${PROJECT_NAME} ${${PROJECT_NAME}_sources})
--- a/rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
+++ b/rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
@@ -147,12 +147,9 @@ rcl_logging_ret_t rcl_logging_external_initialize(const char * config_file, rcut
       // logging.
       return RC_LOGGING_RET_ERROR;
     }
-    // We only want the basename of the program name (in case the OS returns a
-    // a full path).
-    char *basename = ::basename(basec);
 
     char log_name_buffer[512] = {0};
-    snprintf(log_name_buffer, sizeof(log_name_buffer), "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir, basename, rcutils_get_pid(), ms_since_epoch);
+    snprintf(log_name_buffer, sizeof(log_name_buffer), "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir, basec, rcutils_get_pid(), ms_since_epoch);
     allocator.deallocate(basec, allocator.state);
     fprintf(stderr, "CHRIS: log_name_buffer: %s\n", log_name_buffer);
     std::string log_name_str(log_name_buffer);
--- a/rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
+++ b/rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
@@ -26,7 +26,7 @@
 
 #include <rcutils/allocator.h>
 #include <rcutils/get_env.h>
-#include <rcutils/program.h>
+#include <rcutils/process.h>
 #include <rcutils/time.h>
 
 /**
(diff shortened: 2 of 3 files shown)


dirk-thomas:
Please use `~/.ros/log` as the default base path for any logging (same as in [ROS 2 launch](https://github.com/ros2/launch/blob/cdf29248aa4a8ed7d17f3f615453fe8b60b7dc9d/launch/launch/logging.py#L418) as well as ROS 1). Within that directory please choose a scheme which allows multiple log files from multiple processes to write their data into different files.

clalancette:
@nburek We're going to look at changing this to use `~/.ros/log` today to be more like ROS 1.  I'll update this PR with any patches/updates as we make the change.  Let me know if you have any questions or input.

clalancette:
Update:

I have a prototype that puts logs in `~/.ros/log/<exe>-<pid>-<ms_since_epoch>.log`, which is the format used in ROS 1.  I've tested that prototype to work well on Linux.

As part of the prototype, I needed an API change to the external logger initialization so I could allocate memory.  That is now in as https://github.com/ros2/rcl_logging/pull/10 and https://github.com/ros2/rcl/pull/430 .

There is also a need to add some rcutils methods to support things like getting the home directory, executable name, etc. in a cross-platform way.  But since those are strictly API additions, they can come later.

Tomorrow I'll do some additional cleanup on the prototype, test it on Windows and macOS, and then open the PR to rcutils and update this PR with my latest code.

nburek:
I can't approve since I created the original PR, but the changes look good to me.

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -123,8 +117,40 @@ rcl_logging_ret_t rcl_logging_external_initialize(const char * config_file, rcut
...
+    // Get the program name.
+    char *basec = rcutils_get_executable_name(allocator);
+    if (basec == NULL) {
+      // We couldn't get the program name, so get out of here without setting up
+      // logging.
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    char log_name_buffer[512] = {0};
+    snprintf(log_name_buffer, sizeof(log_name_buffer), "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir, basec, rcutils_get_pid(), ms_since_epoch);
```
wjwwood: This line looks really long, is there any linting happening in this file?

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -123,8 +117,40 @@ rcl_logging_ret_t rcl_logging_external_initialize(const char * config_file, rcut
...
+    rcutils_ret_t ret = rcutils_system_time_now(&now);
+    if (ret != RCUTILS_RET_OK) {
+      // We couldn't get the system time, so get out of here without setting up
+      // logging.
+      return RC_LOGGING_RET_ERROR;
+    }
+    int64_t ms_since_epoch = RCUTILS_NS_TO_MS(now);
+
+    // Get the program name.
+    char *basec = rcutils_get_executable_name(allocator);
```
wjwwood: Just curious what does `basec` mean? Why not `executable_name`?

rcl_logging_log4cxx/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ include_directories(${ament_INCLUDE_DIRS} include)
 if(NOT WIN32)
   add_compile_options(-Wall -Wextra -Wpedantic)
 else()
-  # Disable ddl-interface warnings for STL from log4cxx code, because we are already 
+  # Disable ddl-interface warnings for STL from log4cxx code, because we are already
```
wjwwood: `dll`.

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
clalancette: That's true.  The linter isn't enabled on this package.  I'll open an issue to track turning on the linter and dealing with the fallout.

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
clalancette: It is short for "basename copy", but I agree that `executable_name` would be a better term.  I'll do a followup PR tomorrow to rename this.

nuclearsandwich: > I agree that `executable_name` would be a better term. I'll do a followup PR tomorrow to rename this.

I've already got this WIP.

dirk-thomas:
This patch might be responsible for this build failure: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/45/consoleFull#console-section-286 Please double check and address the problem asap. Thanks.

nburek:
> This patch might be responsible for this build failure: https://ci.ros2.org/view/nightly/job/nightly_linux_clang_libcxx/45/consoleFull#console-section-286 Please double check and address the problem asap. Thanks.

I started looking into this and it appears that the issue is that log4cxx isn't able to link when you compile with libcxx. I suspect this is because log4cxx is exposing std library objects in their API and these are different between the std lib that log4cxx was compiled with and libcxx.

dirk-thomas:
@emersonknapp FYI

nburek:
I attempted to get log4cxx building with libc++ to see if we could do a build from source for this nightly CI job or switch to using a vendor package for log4cxx, but am running into issues getting the Apache Portable Runtime (APR) to compile with libc++. 

After talking with @emersonknapp, we think that the quickest way to unblock this nightly CI is to swap the logging implementation using the "RCL_LOGGING_IMPLEMENTATION" environment variable so that it isn't trying to link against log4cxx. We can open a Github issue to track supporting libc++ for rcl_logging_log4cxx if that's something we decide is worth doing. @dirk-thomas Do you agree with this approach?

dirk-thomas:
Sounds fine with me.

nburek:
I've started a PR to fix the nightly build: https://github.com/ros2/ci/pull/291

PR #12, opened 2019-05-06
Followup from #9 and enable linters.

### LINKED ISSUE #11: Enable the linter on both rcl_logging_log4cxx and rcl_logging_noop
Followup from https://github.com/ros2/rcl_logging/pull/9#discussion_r281342407.  This will make the packages conform to the ROS 2 style a bit more.  There will probably be some code changes required to make the linter pass.

nuclearsandwich:
This is a followup from #9 and the start of a resolution for #11 (enables linters but does not address any of the reported issues).

ament_copyright fails on the included FindLog4cxx.cmake file. I don't see FindLog4cxx.cmake in the current HEAD of the [CMake upstream repository](https://gitlab.kitware.com/cmake/cmake) nor have I seen a way to disable copyright checks for just one file while keeping them for the rest of the package.

Fixes #11


nuclearsandwich:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6995/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3245/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5738/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6816/)

clalancette:

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8281/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4234/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6750/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8170/)

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -41,32 +44,19 @@ static const log4cxx::LoggerPtr get_logger(const char * name)
...
-    implementations to map between the incoming integer severity from ROS to the concept of DEBUG, INFO, WARN, ERROR,
-    and FATAL*/
-enum RC_LOGGING_LOG_SEVERITY
-{
-  RC_LOGGING_SEVERITY_UNSET = 0,  ///< The unset log level
-  RC_LOGGING_SEVERITY_DEBUG = 10,  ///< The debug log level
-  RC_LOGGING_SEVERITY_INFO = 20,  ///< The info log level
-  RC_LOGGING_SEVERITY_WARN = 30,  ///< The warn log level
-  RC_LOGGING_SEVERITY_ERROR = 40,  ///< The error log level
-  RC_LOGGING_SEVERITY_FATAL = 50,  ///< The fatal log level
```
sloretz: Were these used anywhere downstream?

clalancette: No, they are only used internally (especially because they were only defined in the `.cpp` file).  They are actually duplicates of the [RCUTILS_LOG_SEVERITY enum](https://github.com/ros2/rcutils/blob/3daf12617c2bfe7d6eb49c68e2710d7727ef2e6d/include/rcutils/logging.h#L164), which is why I wanted to get rid of them.

(also, we never used this logging implementation in a release)

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -78,37 +68,37 @@ extern "C" {
...
 #include "rcl_logging_log4cxx/logging_interface.h"
 
-#define RC_LOGGING_RET_OK                          (0)
-#define RC_LOGGING_RET_WARN                        (1)
-#define RC_LOGGING_RET_ERROR                       (2)
-#define RC_LOGGING_RET_INVALID_ARGUMENT            (11)
-#define RC_LOGGING_RET_CONFIG_FILE_DOESNT_EXIST    (21)
-#define RC_LOGGING_RET_CONFIG_FILE_INVALID         (22)
+#define RCL_LOGGING_RET_OK                          (0)
+#define RCL_LOGGING_RET_ERROR                       (2)
```
sloretz: It seems odd that this is defined in a `cpp` file. What's the relation of the functions in this file with https://github.com/ros2/rcl/blob/master/rcl/include/rcl/logging_external_interface.h ? If the return type is [rcl_ret_t](https://github.com/ros2/rcl/blob/d655147310e3e63aff1555ad40fca1bb9cb7783e/rcl/include/rcl/logging_external_interface.h#L41), then it seems like the return codes should be in https://github.com/ros2/rcl/blob/master/rcl/include/rcl/types.h .

clalancette: Sigh, yes.  This entire package hasn't exactly been done in the correct way.

The relationship with `logging_external_interface.h` is that it *must* implement the exact same API.  But we have to essentially hand-duplicate that API interface here to break a cycle.  The `rcl` package depends on the default implementation (which we are currently switching to `rcl_logging_spdlog`), so the implementation cannot also depend on `rcl` in order to get access to the API header.

Probably the correct answer is to split that API header out of `rcl` and either into its own package or into a "common" API package here, but it fell off the radar.

sloretz: Roger. Is there a ticket for moving the API to a separate package?

clalancette: No, but that's a good idea.  I've just opened https://github.com/ros2/rcl_logging/issues/20 to track it.

PR #17, opened 2019-06-28
Implement a backend logger using spdlog.

clalancette:
Along with ros2/rcutils#166 and ros2/rcl#466, this PR is one way we could implement per-process logging in ROS 2.  It uses a header-only C++ library called [spdlog](https://github.com/gabime/spdlog) instead of log4cxx (which has a bunch of problems).  I'm not at all tied to using this library, but it is relatively straightforward and has most of the features we want.  The open questions with this PR (in rough order of priority):

1.  Is spdlog good enough for our purposes?  It supports all of the features of log4cxx that we were using, with the exception of external configuration.  It just doesn't have a way to be configured from a file.  We could define something, we could ignore the feature, or we could find another C++ logging library that has that feature.
1.  When running a simple executable with this in place (`ros2 run demo_nodes_cpp talker`), you get 2 things in `~/.ros/log`; a directory of `<date>-<hostname>-<pid>`, and a log file of `<processname>-<pid>-<ms-since-epoch>`.  This PR creates the latter; I'm not sure where the former is coming from.  We should figure out our strategy here and only make one of them show up.
1.  Assuming we want to use spdlog, how do we want to vendor it?  Here I've copied it into the repository with a LICENSE file (and made the license of the package match), but we could also `ExternalProject` it or do `spdlog_vendor` package separately.
1.  Needs to be tested with Windows, macOS, etc.  I'll hold off on this until the points above are addressed.

@ros2/team @nburek FYI.

### DIFF
--- a/rcl_logging_spdlog/package.xml
+++ b/rcl_logging_spdlog/package.xml
@@ -6,7 +6,7 @@
   <description>C API providing common interface to a shared library wrapping 3rd party loggers.</description>
   <maintainer email="clalancette@openrobotics.org">Chris Lalancette</maintainer>
   <author email="clalancette@openrobotics.org">Chris Lalancette</author>
-  <license>MIT</license>
+  <license>Apache License 2.0</license>
 
   <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
--- a/rcl_logging_spdlog/package.xml
+++ b/rcl_logging_spdlog/package.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0"?>
-<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<?xml-model href="http://download.ros.org/schema/package_format3.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
 <package format="3">
   <name>rcl_logging_spdlog</name>
-  <version>0.1.0</version>
-  <description>C API providing common interface to a shared library wrapping 3rd party loggers.</description>
+  <version>0.3.0</version>
+  <description>Implementation of rcl_logging API for an spdlog backend.</description>
   <maintainer email="clalancette@openrobotics.org">Chris Lalancette</maintainer>
-  <author email="clalancette@openrobotics.org">Chris Lalancette</author>
   <license>Apache License 2.0</license>
+  <author email="clalancette@openrobotics.org">Chris Lalancette</author>
 
   <buildtool_depend>ament_cmake_ros</buildtool_depend>
 
@@ -14,6 +14,9 @@
   <build_depend>spdlog</build_depend>
 
   <depend>rcutils</depend>
+
+  <test_depend>ament_lint_auto</test_depend>
+  <test_depend>ament_lint_common</test_depend>
 
   <member_of_group>rcl_logging_packages</member_of_group>
 
--- a/rcl_logging_spdlog/package.xml
+++ b/rcl_logging_spdlog/package.xml
@@ -1,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rcl_logging_spdlog</name>
+  <version>0.1.0</version>
+  <description>C API providing common interface to a shared library wrapping 3rd party loggers.</description>
+  <maintainer email="clalancette@openrobotics.org">Chris Lalancette</maintainer>
+  <author email="clalancette@openrobotics.org">Chris Lalancette</author>
+  <license>MIT</license>
+
+  <buildtool_depend>ament_cmake_ros</buildtool_depend>
+
+  <build_depend>spdlog_vendor</build_depend>
+  <build_depend>spdlog</build_depend>
+
+  <depend>rcutils</depend>
+
+  <member_of_group>rcl_logging_packages</member_of_group>
+
+  <export>
+    <build_type>ament_cmake</build_type>
+  </export>
+</package>
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -1,0 +1,155 @@
+// Copyright 2019 Open Source Robotics Foundation, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cerrno>
+#include <inttypes.h>
+
+#include <rcutils/allocator.h>
+#include <rcutils/filesystem.h>
+#include <rcutils/get_env.h>
+#include <rcutils/process.h>
+#include <rcutils/time.h>
+
+#include "spdlog/spdlog.h"
+#include "spdlog/sinks/basic_file_sink.h"
+
+#include "rcl_logging_spdlog/logging_interface.h"
+
+#define RC_LOGGING_RET_OK                          (0)
+#define RC_LOGGING_RET_ERROR                       (2)
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+static std::shared_ptr<spdlog::logger> g_root_logger;
+
+/* These are defined here to match the severity levels in rcl. They provide a consistent way for external logger
+    implementations to map between the incoming integer severity from ROS to the concept of DEBUG, INFO, WARN, ERROR,
+    and FATAL*/
+enum RC_LOGGING_LOG_SEVERITY
+{
+  RC_LOGGING_SEVERITY_UNSET = 0,  ///< The unset log level
+  RC_LOGGING_SEVERITY_DEBUG = 10,  ///< The debug log level
+  RC_LOGGING_SEVERITY_INFO = 20,  ///< The info log level
+  RC_LOGGING_SEVERITY_WARN = 30,  ///< The warn log level
+  RC_LOGGING_SEVERITY_ERROR = 40,  ///< The error log level
+  RC_LOGGING_SEVERITY_FATAL = 50,  ///< The fatal log level
+};
+
+static spdlog::level::level_enum map_external_log_level_to_library_level(int external_level)
+{
+  spdlog::level::level_enum level;
+
+  // map to the next highest level of severity
+  if (external_level <= RC_LOGGING_SEVERITY_DEBUG) {
+    level = spdlog::level::level_enum::debug;
+  } else if (external_level <= RC_LOGGING_SEVERITY_INFO) {
+    level = spdlog::level::level_enum::info;
+  } else if (external_level <= RC_LOGGING_SEVERITY_WARN) {
+    level = spdlog::level::level_enum::warn;
+  } else if (external_level <= RC_LOGGING_SEVERITY_ERROR) {
+    level = spdlog::level::level_enum::err;
+  } else if (external_level <= RC_LOGGING_SEVERITY_FATAL) {
+    level = spdlog::level::level_enum::critical;
+  }
+  return level;
+}
+
+rcl_logging_ret_t rcl_logging_external_initialize(const char * config_file, rcutils_allocator_t allocator)
+{
+  bool config_file_provided = (nullptr != config_file) && (config_file[0] != '\0');
+  if (config_file_provided) {
+    // TODO(clalancette): implement support for an external configuration file.
+    return RC_LOGGING_RET_ERROR;
+  } else {
+    // To be compatible with ROS 1, we want to construct a default filename of
+    // the form ~/.ros/log/<exe>_<pid>_<milliseconds-since-epoch>.log
+
+    // First get the home directory.
+    const char *homedir = rcutils_get_home_dir();
+    if (homedir == NULL) {
+      // We couldn't get the home directory; it is not really going to be
+      // possible to do logging properly, so get out of here without setting
+      // up logging.
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    // SPDLOG doesn't automatically create the log directories, so make them
+    // by hand here.
+    char dotrosdir[512] = {0};
+    snprintf(dotrosdir, sizeof(dotrosdir), "%s/.ros", homedir);
+    if (!rcutils_mkdir(dotrosdir)) {
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    char logdir[512] = {0};
+    snprintf(logdir, sizeof(logdir), "%s/log", dotrosdir);
+    if (!rcutils_mkdir(logdir)) {
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    // Now get the milliseconds since the epoch in the local timezone.
+    rcutils_time_point_value_t now;
+    rcutils_ret_t ret = rcutils_system_time_now(&now);
+    if (ret != RCUTILS_RET_OK) {
+      // We couldn't get the system time, so get out of here without setting up
+      // logging.
+      return RC_LOGGING_RET_ERROR;
+    }
+    int64_t ms_since_epoch = RCUTILS_NS_TO_MS(now);
+
+    // Get the program name.
+    char *basec = rcutils_get_executable_name(allocator);
+    if (basec == NULL) {
+      // We couldn't get the program name, so get out of here without setting up
+      // logging.
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    char log_name_buffer[512] = {0};
+    snprintf(log_name_buffer, sizeof(log_name_buffer), "%s/%s_%i_%" PRId64 ".log", logdir, basec, rcutils_get_pid(), ms_since_epoch);
+    allocator.deallocate(basec, allocator.state);
+
+    g_root_logger = spdlog::basic_logger_mt("root", log_name_buffer);
+    g_root_logger->set_pattern("%v");
+  }
+
+  return RC_LOGGING_RET_OK;
+}
+
+rcl_logging_ret_t rcl_logging_external_shutdown()
+{
+  g_root_logger = nullptr;
+  return RC_LOGGING_RET_OK;
+}
+
+void rcl_logging_external_log(int severity, const char * name, const char * msg)
+{
+  (void)name;
+  g_root_logger->log(map_external_log_level_to_library_level(severity), msg);
+}
+
+rcl_logging_ret_t rcl_logging_external_set_logger_level(const char * name, int level)
+{
+  (void)name;
+
+  g_root_logger->set_level(map_external_log_level_to_library_level(level));
+
+  return RC_LOGGING_RET_OK;
+}
+
+#ifdef __cplusplus
+} /* extern "C" */
+#endif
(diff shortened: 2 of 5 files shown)


jacobperron:
Regarding question 3, can we use package managers for installation:

- [apt](https://packages.ubuntu.com/source/bionic/spdlog) (Linux)
- [brew](https://formulae.brew.sh/formula/spdlog) (macOS)
- [nuget](https://www.nuget.org/packages/spdlog.native/) (Windows) ?

Although, I guess it's possible to run into issues due to different versions.

artivis:
From [spdlog](https://github.com/gabime/spdlog)'s readme, the supported platforms are,
- Linux, FreeBSD, OpenBSD, Solaris, AIX
- Windows (msvc 2013+, cygwin)
- macOS (clang 3.5+)
- Android

while the package managers are,

- Homebrew: brew install spdlog
- FreeBSD: cd /usr/ports/devel/spdlog/ && make install clean
- Fedora: yum install spdlog
- Gentoo: emerge dev-libs/spdlog
- Arch Linux: yaourt -S spdlog-git
- vcpkg: vcpkg install spdlog

I did encounter versioning issues a while back with `spdlog`, hopefully this was smoothed.

fujitatomoya:
I'd like to hear what everyone thinks as following questions,

- sustainability and maintenance.
  need to re-build rcl_logging_spdlog every single time if any fix available on [spdlog](https://github.com/gabime/spdlog)?
- log directory, should be configurable for each process? (or system?)
  with specific and particular situation for debugging, we some time to ramfs.
- from [spdlog](https://github.com/gabime/spdlog) implementation, always use mutex_lock to exclude logger access.
  i believe that nobody like to use the CPU consumption for system but user space.
  there could be buffer to enhance using user space instruction to exclude the access for logger with multi-thread. (example, 1st tries to lock with CAS instruction to exclude the access)

nburek:
My opinions for your question, @clalancette 
1. Yes, I think that spdlog is good enough for our initial purposes. I did some looking around before selecting Log4cxx initially and hadn't discovered any other logging libraries I think are better to continue with than spdlog.  
I also think it's fine to ignore the configuration file input for now. I don't think we need to solve every use case for logging right off the bat, but it would be great just to have something working that dumps logs to file with some sane default settings. That's a big missing feature right now with the dashing release since the log4cxx issues prevented us from getting that enabled by default. Let's let the community drive what additional features they want from logging and then we can explore if we're better off adding a simple config file parser on top of spdlog or if we need to search for a different logger. 
2. I think this second folder in the logs is coming from the ros2/launch system. See: https://github.com/ros2/launch/blob/f1dc16de491864fc2c2a881040003a5f4b1bf8b2/launch/launch/logging/__init__.py#L367-L385. Agreed we should standardize. 
3. Having a copy in the repo was fine for testing that it works, but I don't think it makes sense for that to be the long term solution. For log4cxx we tried used apt, brew, and Chocolatey. The apt and brew repos were already available, but creating the chocolatey package was a real pain and I wouldn't recommend it as a solution for Windows. I personally don't have a strong preference as to how it's vended separately though, so I would defer to others.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,154 @@
...
+    level = spdlog::level::level_enum::critical;
+  }
+  return level;
+}
+
+rcl_logging_ret_t rcl_logging_external_initialize(const char * config_file, rcutils_allocator_t allocator)
+{
+  bool config_file_provided = (nullptr != config_file) && (config_file[0] != '\0');
+  if (config_file_provided) {
+    // TODO(clalancette): deal with the config file
```
nburek: We could return some sort of NOT_IMPLEMENTED error code here so that it fails if someone tries to provide a config, but I think it'd be just as good to remove this if block for now and just mark the config_file as unused.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // SPDLOG doesn't automatically create the log directories, so make them
+    // by hand here.
+    char dotrosdir[512] = {0};
+    snprintf(dotrosdir, sizeof(dotrosdir), "%s/.ros", homedir);
+    if (!rcutils_mkdir(dotrosdir)) {
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    char logdir[512] = {0};
+    snprintf(logdir, sizeof(logdir), "%s/log", dotrosdir);
```
nburek: Check the return code of `snprintf` to make sure that it is `>0` and `<sizeof(logdir)`. 

You probably also want to use the snprintf function defined by rcutils for slightly better cross platform support: https://github.com/ros2/rcutils/blob/e52f2cfc1aac935bb0e6c35a7f3ad332a56d6d71/include/rcutils/snprintf.h#L57

Same comments for the usage of `snprintf` below as well.

clalancette:
>     * sustainability and maintenance.
>       need to re-build rcl_logging_spdlog every single time if any fix available on [spdlog](https://github.com/gabime/spdlog)?

Yeah, that will be a problem with vendoring (either here in this package, or in a vendor package).  So if we can (and as pointed out by others), it would be better to use packages and offload that work elsewhere.

> 
>     * log directory, should be configurable for each process? (or system?)
>       with specific and particular situation for debugging, we some time to ramfs.

Yeah, that's a good point.  That is one reason to support the external configuration file, but on the the other hand, we can just add a new environment variable for that.  I'll put it on the list to do.

> 
>     * from [spdlog](https://github.com/gabime/spdlog) implementation, always use mutex_lock to exclude logger access.
>       i believe that nobody like to use the CPU consumption for system but user space.
>       there could be buffer to enhance using user space instruction to exclude the access for logger with multi-thread. (example, 1st tries to lock with CAS instruction to exclude the access)

I'm not going to worry too much about this unless/until someone comes up with a realistic scenario where this is a problem.  If it does turn out to be a problem, then we can either enhance spdlog upstream, or switch to yet another logging implementation.

Thanks for the feedback!

clalancette:
>     1. Yes, I think that spdlog is good enough for our initial purposes. I did some looking around before selecting Log4cxx initially and hadn't discovered any other logging libraries I think are better to continue with than spdlog.
>        I also think it's fine to ignore the configuration file input for now. I don't think we need to solve every use case for logging right off the bat, but it would be great just to have something working that dumps logs to file with some sane default settings. That's a big missing feature right now with the dashing release since the log4cxx issues prevented us from getting that enabled by default. Let's let the community drive what additional features they want from logging and then we can explore if we're better off adding a simple config file parser on top of spdlog or if we need to search for a different logger.

OK, that was my feeling as well.  I just wanted to make sure there wasn't some critical use case I was missing.

> 
>     2. I think this second folder in the logs is coming from the ros2/launch system. See: https://github.com/ros2/launch/blob/f1dc16de491864fc2c2a881040003a5f4b1bf8b2/launch/launch/logging/__init__.py#L367-L385. Agreed we should standardize.

What confused me was that I wasn't using the launch system.  That directory is showing up even when I just do `ros2 run demo_nodes_cpp talker` with this spdlog implementation enabled.  Maybe somehow `ros2 run` re-uses some of that code?  Not sure.

> 
>     3. Having a copy in the repo was fine for testing that it works, but I don't think it makes sense for that to be the long term solution. For log4cxx we tried used apt, brew, and Chocolatey. The apt and brew repos were already available, but creating the chocolatey package was a real pain and I wouldn't recommend it as a solution for Windows. I personally don't have a strong preference as to how it's vended separately though, so I would defer to others.

OK, I think it is clear that we want to use separate packages.  It seems to already be packaged for Ubuntu, and available via Brew for macOS, so that's a good start.  I'll leave it to @nuclearsandwich and @mjcarroll to give any feedback about chocolatey packaging (or whatever) for Windows.

fujitatomoya:
@clalancette 

> What confused me was that I wasn't using the launch system. That directory is showing up even when I just do ros2 run demo_nodes_cpp talker with this spdlog implementation enabled. Maybe somehow ros2 run re-uses some of that code? Not sure.

this is done by luanch.logging, ros2cli imports logging,

https://github.com/ros2/launch/blob/40c9d879ae01b7c689438461338458b8173e535a/launch/launch/logging/__init__.py#L402-L404

nuclearsandwich:
>     3\. Having a copy in the repo was fine for testing that it works, but I don't think it makes sense for that to be the long term solution. For log4cxx we tried used apt, brew, and Chocolatey. The apt and brew repos were already available, but creating the chocolatey package was a real pain and I wouldn't recommend it as a solution for Windows. I personally don't have a strong preference as to how it's vended separately though, so I would defer to others.

log4cxx definitely set a new worst case bar for Chocolatey packaging.
spdlog has a few advantages that log4cxx didn't.
1. It's a header-only library which makes packaging much easier.
2. Unlike log4cxx a spdlog package already exists for vcpkg. Our CI workflow isn't compatible with vcpkg directly but there's some work in progress that may make extracting chocolatey packages from vpkg possible. https://github.com/microsoft/vcpkg/pull/6891 Testing with spdlog may provide some good practical feedback for that effort.

nuclearsandwich:
@clalancette it wasn't immediately clear which version of spdlog is vendored here. Ubuntu bionic packages 0.16 but the current version in Homebrew is 1.3.1.

clalancette:
This is still on my list to come back and look at for Eloquent, but I haven't had time for it yet.  Probably September.

clalancette:
Looking at the above list of platforms, I think it is broadly clear that we want to use spdlog 1.3.1, as that will get us compatibility on Buster and macOS/brew out-of-the-box.  For Ubuntu Bionic and Windows, then, I think we should probably vendor it like we do in https://github.com/ros2/console_bridge_vendor/blob/master/CMakeLists.txt .  Essentially, if we find spdlog 1.3.1 on the system, use that, otherwise vendor it ourselves.  I'm going to start working on that.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: Good call about the use of snprintf.  I've switch to `rcutils_snprintf` everywhere and now check for errors.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: I'm now using the `RC_LOGGING_RET_ERROR` return here as a stand-in for "this isn't supported yet".  That's on the theory that it is better to fail fast rather than to silently ignore what the user asked for.  So I'm going to mark this as resolved for now, but feel free to make a new comment if you feel we should do something different here.

clalancette:
CI is definitely going to fail here because it needs the whole set of things in https://github.com/ros2/ros2/issues/789 in order to succeed.

rcl_logging_spdlog/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rcl_logging_spdlog</name>
+  <version>0.1.0</version>
+  <description>C API providing common interface to a shared library wrapping 3rd party loggers.</description>
+  <maintainer email="clalancette@openrobotics.org">Chris Lalancette</maintainer>
+  <author email="clalancette@openrobotics.org">Chris Lalancette</author>
+  <license>MIT</license>
```
jacobperron: The files added have an Apache 2.0 license.
```suggestion
  <license>Apache License 2.0</license>
```

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <rcutils/snprintf.h>
+#include <rcutils/time.h>
+
+#include "spdlog/spdlog.h"
+#include "spdlog/sinks/basic_file_sink.h"
+
+#include "rcl_logging_spdlog/logging_interface.h"
+
+#define RC_LOGGING_RET_OK                          (0)
+#define RC_LOGGING_RET_ERROR                       (2)
```
jacobperron: Are the parens necessary? Also the white space seems excessive.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,174 @@
...
+extern "C" {
+#endif
+
+static std::mutex g_logger_mutex;
+static std::shared_ptr<spdlog::logger> g_root_logger = nullptr;
+
+/* These are defined here to match the severity levels in rcl. They provide a consistent way for external logger
+    implementations to map between the incoming integer severity from ROS to the concept of DEBUG, INFO, WARN, ERROR,
+    and FATAL*/
+enum RC_LOGGING_LOG_SEVERITY
```
jacobperron: Could we use the rcutils log levels instead of redefining here?

https://github.com/ros2/rcutils/blob/54ecb72a3784cc54d361f38bb9cab5bad7172772/include/rcutils/logging.h#L164

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,174 @@
...
+  // the tests do this implicitly by calling rclcpp::init more than once).
+  // If the logger is already setup, don't do anything.
+  if (g_root_logger != nullptr) {
+    return RC_LOGGING_RET_OK;
+  }
+
+  bool config_file_provided = (nullptr != config_file) && (config_file[0] != '\0');
+  if (config_file_provided) {
+    // TODO(clalancette): implement support for an external configuration file.
+    return RC_LOGGING_RET_ERROR;
```
jacobperron: Does it make sense to set an error message with `RCUTILS_SET_ERROR_MSG` before returning? Same for other error cases below.

rcl_logging_spdlog/package.xml
clalancette: Good call, will fix.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: They are not strictly necessary, but are usually recommended when doing macros like this.  I'll fix up the whitespace.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: Oh, good call.  Will change to that.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: Yeah, that makes sense.  I just didn't know about RCUTILS_SET_ERROR_MSG.  Will fix.

clalancette:
Thanks for the initial review!  All review comments have been addressed in b32968f.  Please take another look when you have a chance.

clalancette:
> Ultimately, it looks like we might want to define a common logging API so that the backend can be easily changed. But I'm okay with this PR as-is, if we want to leave that as a follow-up task.

So there actually is one that they all conform to: https://github.com/ros2/rcl/blob/d655147310e3e63aff1555ad40fca1bb9cb7783e/rcl/include/rcl/logging_external_interface.h .  It's a little hard to find it, though, because of the way this whole thing was done.  I honestly don't remember all of the details at the moment, but it definitely could be improved so that they use more common stuff.

rcl_logging_spdlog/package.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="3">
+  <name>rcl_logging_spdlog</name>
+  <version>0.1.0</version>
+  <description>C API providing common interface to a shared library wrapping 3rd party loggers.</description>
```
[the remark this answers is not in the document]
clalancette: Yeah, good call.  Will fix.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,171 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <cerrno>
+#include <memory>
+#include <mutex>
+#include <inttypes.h>
```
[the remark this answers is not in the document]
clalancette: Done and done.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include <rcutils/snprintf.h>
+#include <rcutils/time.h>
+
+#include "spdlog/spdlog.h"
+#include "spdlog/sinks/basic_file_sink.h"
+
+#include "rcl_logging_spdlog/logging_interface.h"
+
+#define RC_LOGGING_RET_OK    (0)
+#define RC_LOGGING_RET_ERROR (2)
```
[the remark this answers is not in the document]
clalancette: As we see in https://github.com/ros2/rcl/blob/d655147310e3e63aff1555ad40fca1bb9cb7783e/rcl/include/rcl/logging_external_interface.h , the return value for these methods is actually `rcl_ret_t`.  The reason we duplicate them here (as opposed to use the RCL_RET_OK, etc return values) is to make sure we don't have a circular dependency between this package and rcl (which specifies one of these backends as the default).  So I don't think we need to export these; callers will end up using the `rcl_ret_t` values instead.

I agree with you in making them start with `RCL`, though.  I'll do that.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -0,0 +1,171 @@
...
+    // Get the program name.
+    char *basec = rcutils_get_executable_name(allocator);
+    if (basec == NULL) {
+      // We couldn't get the program name, so get out of here without setting up
+      // logging.
+      RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
+      return RC_LOGGING_RET_ERROR;
+    }
+
+    print_ret = rcutils_snprintf(name_buffer, sizeof(name_buffer), "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir, basec, rcutils_get_pid(), ms_since_epoch);
```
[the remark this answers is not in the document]
clalancette: Yeah, I'll break it up.

clalancette:
All review comments are addressed in e268e69.

nuclearsandwich:
The Epr job won't work until spdlog_vendor is released into eloquent. I think it would be nice if we got the PR job passing before merging this.

clalancette:

clalancette:

clalancette:
Woohoo, PR testing succeeded.  Merging now.

pwm1234:
> 1. Is spdlog good enough for our purposes?  It supports all of the features of log4cxx that we were using, with the exception of external configuration.  It just doesn't have a way to be configured from a file.  We could define something, we could ignore the feature, or we could find another C++ logging library that has that feature.

@clalancette For reference https://github.com/guangie88/spdlog_setup is a reasonable place to start when it comes to configuring spdlog from a file. I have used it on a few small projects and have been pleased.

maxlein:
Is there somewhere a hint or manual on how to switch to spdlog?  
I asked this in ros answers few months ago but didn‘t get an answer.

nburek:
I believe that as of the Eloquent release spdlog is now the default logger, someone correct me if I'm wrong. If you're on a previous version, you will need to build rcl_logging_spdlog into your workspace (if it isn't already available) and then you will need to rebuild rcl while linking in rcl_logging_spdlog as the logger you want. You can do this by either copying the changes from [here](https://github.com/ros2/rcl/pull/466/files) to set it as the default logger or by setting the environment variable `RCL_LOGGING_IMPLEMENTATION=rcl_logging_spdlog` during the build so that the make files pick it up ([see here](https://github.com/ros2/rcl/blob/master/rcl/cmake/get_default_rcl_logging_implementation.cmake#L28-L31))

clalancette:
That's correct; in Eloquent and later, spdlog is now the default.

maxlein:
Thanks, I am on eloquent.
So to check, I will look at the env var.

PR #21, opened 2019-10-04
Fix Clang warning about possible uninitialized variable

jacobperron:
See warning: https://ci.ros2.org/job/ci_osx/6754/warnings11Result/

macOS: [Build Status](https://ci.ros2.org/job/ci_osx/6758/)

PR #22, opened 2019-10-06
fix initialized warning

Karsten1987:
I get warnings on OSX for this.
If non of the cases in the `if/else` tree kicks in, the `level` variable is uninitialized. I am up for opinions, I thought it makes sense to default to `info`.

Signed-off-by: Karsten Knese <karsten@openrobotics.org>


dirk-thomas:
Duplicate of #21.

PR #23, opened 2019-12-06
Bypass spdlog singleton registry

ivanpauno:
Fixes ros2/rclcpp#933.

### DIFF
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -24,6 +24,7 @@
 #include <cinttypes>
 #include <memory>
 #include <mutex>
+#include <utility>
 
 #include "spdlog/spdlog.h"
 #include "spdlog/sinks/basic_file_sink.h"
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -38,7 +38,7 @@ extern "C" {
 #endif
 
 static std::mutex g_logger_mutex;
-static std::shared_ptr<spdlog::logger> g_root_logger = nullptr;
+static std::unique_ptr<spdlog::logger> g_root_logger = nullptr;
 
 static spdlog::level::level_enum map_external_log_level_to_library_level(int external_level)
 {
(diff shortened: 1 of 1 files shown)


ivanpauno:
@clalancette @hidmic can you review this?

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -142,7 +142,10 @@ rcl_logging_ret_t rcl_logging_external_initialize(
       RCUTILS_SET_ERROR_MSG("Failed to create log file name string");
       return RCL_LOGGING_RET_ERROR;
     }
-    g_root_logger = spdlog::basic_logger_mt("root", name_buffer);
+
+    auto sink = std::make_shared<spdlog::sinks::basic_file_sink_mt>(name_buffer, false);
```
clalancette: I think we can clarify the ownership semantics here by making this a `std::unique_ptr`.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -152,7 +155,6 @@ rcl_logging_ret_t rcl_logging_external_initialize(
 rcl_logging_ret_t rcl_logging_external_shutdown()
 {
   g_root_logger = nullptr;
```
clalancette: Did you test (with e.g. valgrind) to make sure that the logger is getting cleaned up properly after this patch?  I'm assuming it is, I'm just curious to see the results.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
ivanpauno: Are you referring to the `sink` or the `g_root_logger`?
About the first, the function takes a `shared_ptr`. IMO, creating and moving any of both smart pointers is equally clear (and should produce the same code), but I changed to `unique_ptr`.
About the second, I moved to an `unique_ptr`.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
ivanpauno: I've tested now, it doesn't leak.
I've also double checked the testing procedure with an intentionally leaking version.

clalancette: Fantastic, thanks!

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
clalancette: Looks good, thanks.

ivanpauno:
One full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8844/)

The others up to rcl_logging_spdlog:
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4645/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7246/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8763/)

Issue #26, opened 2020-03-09
missing run dependency on spdlog on Ubuntu Focal


dirk-thomas:
See e.g. http://build.ros2.org/view/Fci/job/Fci__nightly-cross-vendor-connext-cyclonedds_ubuntu_focal_amd64/5/console which fails to build an overlay workspace while `rcl_logging` is part of the underlay.

clalancette:
In ROS 2 Eloquent on Ubuntu Bionic, Windows, and macOS, we use spdlog as a header-only library, so we never exported it as a `exec_depend`.  But it is also the case that we only install the vendored version of the library if version < 1.3.1 is available.  It looks like Focal ships with spdlog version 1.5.0 as a compiled library, so we use the system library there.

For Foxy, this gets us into an interesting situation where we are doing something different for Ubuntu vs Windows/macOS.  We have a few options to fix the problem:

1.  Just add an `exec_depend` in the proper place.  This is the easiest, but also the least consistent.
1.  Always vendor spdlog, even when it is available as a system dependency.  This keeps maximum consistency between our platforms.
1.  Vendor spdlog if it is too old, but build it as a shared library.  This keeps consistency between the platforms.
1.  Build an spdlog package for homebrew and chocolatey, and never vendor spdlog.  Given the rolling nature of homebrew, though, this is likely to cause compatibility problems later.

Personally, I'd go for 3 in the list above, where we build a vendored spdlog as a shared library only if it is too old.  Thoughts?

dirk-thomas:
Option 3 sounds good to me.

clalancette:
So, as usual, Windows is a pain.  We can build spdlog for both Linux and macOS, but spdlog doesn't currently have a way to build the library for Windows.  That means that the only option on Windows is to do header-only.  So what I'm going to suggest here is the following:

1.  We update the documentation for macOS for Foxy to include `brew install spdlog`.  As of right now, that will bring in the binary version of spdlog 1.5.0.
1.  Ubuntu Focal will already bring in the spdlog dependency via the rosdep install line.
1.  When building the spdlog_vendor package on Ubuntu Focal and macOS, it will just use the system version.
1.  When building the spdlog_vendor package on Ubuntu Bionic, it will build the binary version.
1.  When building the spdlog_vendor package on Windows, we will copy the header-only version into place.

clalancette:
CI:
[...]

PR #27, opened 2020-03-11
Make spdlog an exec_depend.

clalancette:
Newer spdlog (>= 1.5.0) builds a shared library, so we
need to add an exec_depend now.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

Part of solving https://github.com/ros2/rcl_logging/issues/26

### DIFF
--- a/rcl_logging_spdlog/CMakeLists.txt
+++ b/rcl_logging_spdlog/CMakeLists.txt
@@ -47,6 +47,6 @@ if(BUILD_TESTING)
 endif()
 
 ament_export_include_directories(include)
-ament_export_dependencies(ament_cmake rcutils)
+ament_export_dependencies(ament_cmake rcutils spdlog_vendor spdlog)
 ament_export_libraries(${PROJECT_NAME})
 ament_package()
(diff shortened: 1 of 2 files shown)


dirk-thomas:
Doesn't it also export the dependency in CMake?

clalancette:
> Doesn't it also export the dependency in CMake?

So I did a full build of the ros2 workspace without this, and it succeeded.  I then tested out some of `demo_nodes_cpp`, which also succeeded.  I also did an overlay build which succeeded.  So I'm not entirely sure if it is needed.

That being said, it is more correct so I'll add it.  Thanks for the heads up.

cottsay:
@clalancette - Can this be backported to Eloquent? I'm getting RPM build failures on Fedora because `libspdlog.so.1` isn't installed because the `spdlog` dependency is missing.

clalancette:
> @clalancette - Can this be backported to Eloquent? I'm getting RPM build failures on Fedora because `libspdlog.so.1` isn't installed because the `spdlog` dependency is missing.

I don't see why not.  It will never be strictly needed on our Eloquent supported platforms, but it shouldn't hurt either.  @cottsay Would you mind opening up the backport PR?  I'll review it.

PR #36, opened 2020-05-14
Add some preliminary functional tests

cottsay:
Eventually we'll want to create a test to verify that we have spdlog configured correctly and can log data, but right now there aren't any good mechanisms to keep us from showering the `~/.ros/log/` directory with our log files. There also isn't a great way to determine exactly where the log ended up.

I'll follow-up with changes to add more comprehensive testing when those mechanisms are available.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10754/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6149/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8756/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10643/)

### DIFF
--- a/rcl_logging_spdlog/test/test_logging_interface.cpp
+++ b/rcl_logging_spdlog/test/test_logging_interface.cpp
@@ -31,11 +31,11 @@ TEST(logging_interface, init_invalid)
   bad_allocator.allocate = bad_malloc;
 
   // Config files are not supported by spdlog
-  EXPECT_NE(0, rcl_logging_external_initialize("anything", allocator));
+  EXPECT_EQ(2, rcl_logging_external_initialize("anything", allocator));
   rcutils_reset_error();
-  EXPECT_NE(0, rcl_logging_external_initialize(nullptr, bad_allocator));
+  EXPECT_EQ(2, rcl_logging_external_initialize(nullptr, bad_allocator));
   rcutils_reset_error();
-  EXPECT_NE(0, rcl_logging_external_initialize(nullptr, invalid_allocator));
+  EXPECT_EQ(2, rcl_logging_external_initialize(nullptr, invalid_allocator));
   rcutils_reset_error();
 }
 
--- a/rcl_logging_spdlog/test/test_logging_interface.cpp
+++ b/rcl_logging_spdlog/test/test_logging_interface.cpp
@@ -14,6 +14,7 @@
 
 #include <rcutils/allocator.h>
 #include <rcutils/error_handling.h>
+#include <rcutils/logging.h>
 #include "rcl_logging_spdlog/logging_interface.h"
 #include "gtest/gtest.h"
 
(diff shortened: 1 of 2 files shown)


rcl_logging_spdlog/test/test_logging_interface.cpp
```diff
@@ -0,0 +1,35 @@
...
+}
+
+TEST(logging, init_invalid)
+{
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  rcutils_allocator_t bad_allocator = rcutils_get_default_allocator();
+  bad_allocator.allocate = bad_malloc;
+
+  // Config files are not supported by spdlog
+  ASSERT_NE(0, rcl_logging_external_initialize("anything", allocator));
```
brawner, line 32: Use explicit error level
```
ASSERT_EQ(RCL_LOGGING_RET_ERROR, ...
```

rcl_logging_spdlog/test/test_logging_interface.cpp
```diff
@@ -0,0 +1,35 @@
...
+  rcutils_reset_error();
+  ASSERT_NE(0, rcl_logging_external_initialize(nullptr, bad_allocator));
```
brawner, line 34: Can you walk through the initialize, log and shutdown steps? No need to be thorough at this moment, but stepping through the typical use cases would be good.

cottsay: Successful initialization results in log files being created. Once the mechanisms I mentioned in the description are merged and I'm able to make sure those logs will be cleaned up, I can add initialization tests (and tests for levels).

brawner: Aren't log files typically generated in tests then, if rcl_init/rcl_init_node are ever called? How are they cleaning up their logs?

cottsay: Fair point, I'd guess they aren't. I'd suppose it isn't a problem for our buildfarms because the containers get wiped out.

I'll add some init tests now, then.

rcl_logging_spdlog/test/test_logging_interface.cpp
cottsay: I can't do that. Those macros are defined in the C++ file, not the header.

The documentation in `rcl` for these functions actually says that the return values should be `RCL_RET_OK` or `RCL_RET_ERROR`, so this implementation doesn't match that.

I doubt that anyone has taken a dependency on the error codes for these functions, but we're past API freeze for Foxy. Thoughts? Fix it (and log4cxx) in G-turtule?

rcl_logging_spdlog/test/test_logging_interface.cpp
cottsay: Added in 755efc2

rcl_logging_spdlog/test/test_logging_interface.cpp
brawner: My primary suggestion was to use ASSERT_EQ and the actual error code that is returned, so I guess the suggestion with your other point would be: 
```
ASSERT_EQ(2, rcl_logging_external_initialize...)
```
If you can't use the actual macro, that's not a big deal to me.

As for whether to change the return codes, you might want to get input about the API fix from a more seasoned maintainer, but I don't see an issue with fixing something to conform to the actually expected API of the larger ecosystem. This package has yet to be bumped to 1.0, so you might be able to squeeze it in.

cottsay: I'm not super jazzed about specifically asserting undocumented behavior in the tests, but you seem confident that it's the right thing to do, so I'll roll with it. Updated in 7d52599.

brawner: Yah, the undocumented behavior isn't great, and you might want to create a follow-up issue. That being said, I like to view tests as another sort of "documentation". If somebody wants to fix the API in the future, the test doesn't silently acquiesce to a new API. It will have to be changed as well, which also means that it will resist unwanted changes in the future. Meanwhile, people can look at the test and see the warts and all of the current API.

I also love being as explicit as possible that the expected error matches the actual error as much as possible. I would add EXPECT_STREQ with rcutil's error message if it also didn't report the file path in that message.

ahcorde:
I added in this PR https://github.com/ros2/rcl_logging/pull/37/files a function to check that the file is written properly

hidmic:
@cottsay this one's ready to go, isn't it? @ahcorde improvements could be turned into a follow-up PR.

cottsay:
> improvements could be turned into a follow-up PR.

Agreed. Some of the rcutils/rcpputils stuff I have lined up will make that better, anyway.

PR #38, opened 2020-05-15
warn about unused return value for set_logger_level

dirk-thomas:
In combination with ros2/rcl#652.


dirk-thomas:
> These implementations return `rcl_logging_ret_t` instead of `rcl_ret_t`?

Sorry, not part of this PR.

PR #40, opened 2020-06-02
More spdlog tests

cottsay:
As suggested in [#36 (comment)](https://github.com/ros2/rcl_logging/pull/36#issuecomment-629144007), some of this approach takes after #37.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11021/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6353/)
* Linux-coverage [Build Status](http://ci.ros2.org/job/ci_linux/11021/cobertura/src_ros2_rcl_logging_rcl_logging_spdlog_src/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8978/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10948/)

PR #41, opened 2020-06-08
Logging interface

clalancette:
As discussed in [several](https://github.com/ros2/rcl_logging/pull/33#issuecomment-629416344) [places](https://github.com/ros2/rcl_logging/pull/40#discussion_r435265222), the current way that the rcl_logging API is working is quite odd.  The API is currently being defined in rcl, implemented in rcl_logging, and then used in rcl.  This essentially creates a circular dependency that we are breaking by copying a bunch of code around.

Instead of that, what this PR (and the counterpart in ros2/rcl#676) does is to define an `rcl_logging_interface` package.  This package *only* installs and provides a header file.  Then the logging backends depend on this interface and implement it.  Finally, rcl also depends on the interface, and calls into whatever backend is setup as the default.  This breaks the circular dependency.

@rotu @wjwwood @cottsay @hidmic FYI.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11047/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6371/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8998/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10970/)

rcl_logging_interface/include/rcl_logging_interface/visibility_control.h
```diff
@@ -0,0 +1,58 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef RCL_LOGGING_INTERFACE__VISIBILITY_CONTROL_H_
+#define RCL_LOGGING_INTERFACE__VISIBILITY_CONTROL_H_
+
+#ifdef __cplusplus
+extern "C"
+{
```
hidmic, line 20: @clalancette nit^2: be consistent with the curly bracket positioning. Some files keep it at the end of the line where `extern "C"` shows, others do this.

clalancette: 04ebe64bc75d3f3dc049e9e9e4b86b3f4f58c5bf should improve the situation.

rcl_logging_noop/CMakeLists.txt
```diff
@@ -46,6 +43,6 @@ install(TARGETS ${PROJECT_NAME}
   RUNTIME DESTINATION bin)
```
cottsay, line 43: Does this `install()` need `EXPORT ${PROJECT_NAME}` like the other two logging implementations?

clalancette: Oh yeah, probably.  Also probably needs `ament_export_target`.  Will add both.

clalancette:
> Thanks, Chris, this is great. Do we need to add a quality declaration since this new package will be in the "critical path"?

Yeah, good point.  Added in 655ee72.

clalancette:
All right, smoke-test CI looked good.  Here is a full CI to make sure nothing wonky happens:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11069/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6382/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9014/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10983/)

clalancette:
All right, Linux warnings are expected and in the nightlies.

macOS failures are all in the nightlies except one.  But given that it is a failure in a service, I think that is probably flaky.

Windows is hard to tell, but there are 4 additional tests that failed here vs. the nightlies.  All of those were in services as well, so I think we can chalk that up to flakiness.

Overall, it looks like this set of PRs doesn't have much impact (which is good!).  Going ahead and merging, thanks for the reviews.

PR #47, opened 2020-07-01
Include rcutils/allocator.h in logging_interface.h

j-rivero:
While running the ABI tool on the package the compilation of the headers displays a problem unable to found [`rcutils_allocator_t` in logging_interface.h](https://github.com/ros2/rcl_logging/compare/foxy...jrivero/missing_allocator_header_foxy?expand=1#diff-09665547ad7117bc6bfca31abb0cf87aR43). I think we are missing the include in the file.

### DIFF
--- a/rcl_logging_spdlog/include/rcl_logging_spdlog/logging_interface.h
+++ b/rcl_logging_spdlog/include/rcl_logging_spdlog/logging_interface.h
@@ -19,6 +19,8 @@
 extern "C" {
 #endif
 
+#include <rcutils/allocator.h>
+
 #include "rcl_logging_spdlog/visibility_control.h"
 
 typedef int rcl_logging_ret_t;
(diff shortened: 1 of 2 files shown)


rcl_logging_spdlog/include/rcl_logging_spdlog/logging_interface.h
```diff
@@ -19,6 +19,8 @@
 extern "C" {
 #endif
 
+#include <rcutils/allocator.h>
```
clalancette, line 22: A couple of nits here:

1.  It would be better to move both of the includes out of the `extern "C"` block; it really isn't necessary to have them in there.
1.  Let's use `#include "rcutils/allocator.h"`, just to be consistent with what is already there.
1.  In sort order, "rcutils/allocator.h" comes after "rcl_logging_spdlog/visibility_control.h".

j-rivero: b9a2e4514598ac4f4ea5011541108f6e7b82c93c

j-rivero:
> Besides the minor changes inline, we should also probably add this include to rcl_logging_log4cxx/include/rcl_logging_log4cxx/logging_interface.h . We don't use it presently, but it is good to keep it up-to-date.

b9a2e4514598ac4f4ea5011541108f6e7b82c93c

> Finally, @j-rivero , you need to put the Signed-off-by line on your commit to pass the DCO bot.

Sorry forgot it. I think it is solved now.

PR #52, opened 2020-09-17
Added benchmark test to rcl_logging_spdlog

ahcorde:
Added benchmark test to `rcl_logging_spdlog`

Signed-off-by: ahcorde <ahcorde@gmail.com>


rcl_logging_spdlog/benchmark/benchmark_logging_interface.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+  RCUTILS_LOG_SEVERITY_UNSET,
+  RCUTILS_LOG_SEVERITY_DEBUG,
+  RCUTILS_LOG_SEVERITY_INFO,
+  RCUTILS_LOG_SEVERITY_WARN,
+  RCUTILS_LOG_SEVERITY_ERROR,
+  RCUTILS_LOG_SEVERITY_FATAL,
+};
+
+class RCL_LOGGING_BENCHMARK : public PerformanceTest
```
cottsay, line 42: Is there a reason this doesn't conform to the [style guide](https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#id3)?

rcl_logging_spdlog/benchmark/benchmark_logging_interface.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+public:
+  void SetUp(benchmark::State & st)
+  {
+    performance_test_fixture::PerformanceTest::SetUp(st);
```
cottsay, line 47: Unless you think it specifically adds clarity, I think you should drop the namespace here. Either that or add it to the class inheritance so that you're consistent.
```suggestion
    PerformanceTest::SetUp(st);
```

rcl_logging_spdlog/benchmark/benchmark_logging_interface.cpp
```diff
@@ -0,0 +1,133 @@
...
+  void TearDown(benchmark::State & st)
+  {
+    performance_test_fixture::PerformanceTest::TearDown(st);
+    rcl_logging_ret_t ret = rcl_logging_external_shutdown();
+    if (ret != RCL_LOGGING_RET_OK) {
+      st.SkipWithError(rcutils_get_error_string().str);
+    }
+  }
+
+  void setLogLevel(int logger_level, benchmark::State & st)
```
cottsay, line 68: For clarity:
```suggestion
  static void setLogLevel(int logger_level, benchmark::State & st)
```

cottsay (changes requested):
I'm getting the sense that you're trying to get good coverage in the performance tests here. I don't think that should be a goal. Arbitrary numbers coming out of the performance test results don't help us figure out what's wrong when they increase, and don't help us make determinations as to whether the performance is "acceptable" because we'd never perform this sort of operation when consuming this API.

I see two scenarios being tested inside of each of these tests: the case where the level of the log operation is a "hit", and the case where it is a "miss" and is ignored. I think we'll be interested in those two cases specifically. I wouldn't expect that to performance to change based on what levels are actually used, only which of those modes occurs.

It would also be good to test the initialize/shutdown sequence.

Nice job with the fixture, though.

ahcorde:
Compiling up-to `rcl_logging_spdlog` and testing `rcl_logging_spdlog` with `-DAMENT_RUN_PERFORMANCE_TESTS=ON`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12419/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7386/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10129/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12332/)

rcl_logging_spdlog/test/benchmark/benchmark_logging_interface.cpp
```diff
@@ -0,0 +1,117 @@
...
+BENCHMARK_DEFINE_F(PerformanceTest, logging_shutdown)(benchmark::State & st)
+{
+  for (auto _ : st) {
+    rcl_logging_ret_t ret = rcl_logging_external_shutdown();
+    if (ret != RCL_LOGGING_RET_OK) {
+      st.SkipWithError(rcutils_get_error_string().str);
+    }
+  }
+}
+BENCHMARK_REGISTER_F(PerformanceTest, logging_shutdown);
```
cottsay, line 117: > It would also be good to test the initialize/shutdown sequence.

I'm sorry this wasn't clear, but I meant to test initialize and shutdown together, so a full cycle. I'm pretty sure subsequent calls to `initialize` over and over will no-op.

```suggestion
BENCHMARK_F(PerformanceTest, logging_initialize_shutdown)(benchmark::State & st)
{
  rcutils_allocator_t allocator = rcutils_get_default_allocator();
  for (auto _ : st) {
    rcl_logging_ret_t ret = rcl_logging_external_initialize(nullptr, allocator);
    if (ret != RCL_LOGGING_RET_OK) {
      st.SkipWithError(rcutils_get_error_string().str);
    }
    ret = rcl_logging_external_shutdown();
    if (ret != RCL_LOGGING_RET_OK) {
      st.SkipWithError(rcutils_get_error_string().str);
    }
  }
}
```

rcl_logging_spdlog/test/benchmark/benchmark_logging_interface.cpp
```diff
@@ -0,0 +1,117 @@
...
+const int logger_levels[] =
+{
+  RCUTILS_LOG_SEVERITY_UNSET,
+  RCUTILS_LOG_SEVERITY_DEBUG,
+  RCUTILS_LOG_SEVERITY_INFO,
+  RCUTILS_LOG_SEVERITY_WARN,
+  RCUTILS_LOG_SEVERITY_ERROR,
+  RCUTILS_LOG_SEVERITY_FATAL,
+};
+
```
cottsay, line 41: This can be dropped now (which should resolve the OSX clang warning CI is complaining about).

```suggestion
```

cottsay (changes requested):
Almost there.

rcl_logging_spdlog/test/benchmark/benchmark_logging_interface.cpp
cottsay: I resolved this with 4bf7cf8.

cottsay:
Fresh builds:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12465/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7432/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10178/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12384/)

cottsay (approved):

PR #53, opened 2020-09-22
Allow configuring logging directory through environment variables

### LINKED ISSUE #50: Configure logging directory
Currently, all logs are saved to the directory `~/.ros/log`, which is hardcoded in the following places:

https://github.com/ros2/rcl_logging/blob/07463919e1a791d7acf69d70f8f06a51b3115319/rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp#L135

https://github.com/ros2/rcl_logging/blob/07463919e1a791d7acf69d70f8f06a51b3115319/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp#L101

https://github.com/ros2/launch/blob/e51484685387d4186358ad128d3a4232a5d06522/launch/launch/logging/__init__.py#L96

In ROS 1, we can change the directory logs are saved to by setting the [`ROS_HOME`](http://wiki.ros.org/ROS/EnvironmentVariables#ROS_HOME) and/or [`ROS_LOG_DIR`](http://wiki.ros.org/ROS/EnvironmentVariables#ROS_LOG_DIR) environment variables.
It would be nice to have a way to configure the directory in ROS 2. 

I think an environment variable is a convenient way for users to configure the logging directory, although I propose using a single environment variable instead of combining two (like in ROS 1).

christophebedard:
This allows configuring the logging directory through environment variables using the following logic:

* Use $ROS_LOG_DIR if ROS_LOG_DIR is set and not empty.
* Otherwise, use $ROS_HOME/log, using ~/.ros for ROS_HOME if not set or if empty.

Since this function is meant to be used by all logging implementations, I put it in `rcl_logging_interface`. I also added a test.

Closes #50

Requires https://github.com/ros2/rcpputils/pull/98


christophebedard:
> Since this function is meant to be used by all logging implementations, I put it in `rcl_logging_interface`.

But maybe it shouldn't be `RCL_LOGGING_INTERFACE_PUBLIC`/in the header directly?

christophebedard:

christophebedard:
oh, build.ros.org doesn't build everything from source.

christophebedard:
@jacobperron could you take a look?

rcl_logging_interface/CMakeLists.txt
```diff
@@ -32,6 +32,21 @@ install(DIRECTORY include/${PROJECT_NAME}/
   DESTINATION include/${PROJECT_NAME}
 )
 
+if(BUILD_TESTING)
```
ivanpauno, line 35: nit^N: It's more usual to see the following order

```cmake
ament_export_*(...)
...

if(BUILD_TESTING)
   ...
endif()

ament_package()
```

though there's no explicit rule about it

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+/// Get the logging directory.
+/**
+ * Uses various environment variables to construct a logging directory path.
+ *
+ * Use $ROS2_LOG_DIR if ROS2_LOG_DIR is set and not empty.
+ * Otherwise, use $ROS2_HOME/log, using ~/.ros for ROS2_HOME if not set or if empty.
+ *
+ * It also expands '~' to the current user's home directory.
+ *
+ * \param[inout] directory The C string pointer at which to write the directory path.
```
ivanpauno, line 107:
```suggestion
 * \param[out] directory The C string pointer at which to write the directory path.
```
We're not reading `directory`, just using it as an output.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ *   Only meaningful if the call is successful.
+ * \param[in] allocator The allocator to use for memory allocation.
```
ivanpauno, line 109: In some places we pass the allocator by value, in some others by pointer.
More often, we do the second so I would change the signature to `const rcutils_allocator_t * allocator)`.

PS: We should really have an item in the style guide clarifying that.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ * \return RCL_LOGGING_RET_OK if successful, or
+ * \return RCL_LOGGING_RET_ERROR if an unspecified error occurs.
+ */
+RCL_LOGGING_INTERFACE_PUBLIC
+inline
+rcl_logging_ret_t
+rcl_logging_get_logging_directory(const char ** directory, rcutils_allocator_t allocator)
```
ivanpauno, line 116:
```suggestion
rcl_logging_get_logging_directory(const rcutils_allocator_t * allocator, const char ** directory)
```

Notes:
- The style guide says that inputs should go first, outputs at the end.
- See comment about allocator by value vs allocator by pointer above.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+{
+  const char * log_dir_env;
+  const char * err = rcutils_get_env("ROS2_LOG_DIR", &log_dir_env);
+  if (!err && *log_dir_env != '\0') {
+    *directory = log_dir_env;
```
ivanpauno, line 121: The output of `rcutils_get_env` must not be freed by the user and the output of `rcutils_join_path` must.
A function cannot return sometimes a borrowed pointer and sometimes an owned pointer, so in this case `log_dir_env` should be duplicated.

The docs should also be clear about ownership of the returned pointer.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+  } else {
+    const char * ros2_home_dir_env;
+    err = rcutils_get_env("ROS2_HOME", &ros2_home_dir_env);
+    if (err || *ros2_home_dir_env == '\0') {
```
ivanpauno, line 125: If the environment variable was not set, `rcutils_get_env` still returns `err = NULL` and `ros2_home_dir_env=""`.
i.e.: if `err` is not `NULL` (error reported), the function should return an error and not try getting a different environment variable.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ * \param[in] allocator The allocator to use for memory allocation.
+ * \return RCL_LOGGING_RET_OK if successful, or
+ * \return RCL_LOGGING_RET_ERROR if an unspecified error occurs.
+ */
+RCL_LOGGING_INTERFACE_PUBLIC
+inline
+rcl_logging_ret_t
+rcl_logging_get_logging_directory(const char ** directory, rcutils_allocator_t allocator)
+{
+  const char * log_dir_env;
```
ivanpauno, line 118: It would be nice to return `RCL_LOGGING_RET_INVALID_ARG` if either:
- `directory` is `NULL` (to avoid segfaults).
- `*directory` is not NULL (to avoid leaks).

rcl_logging_interface/CMakeLists.txt
```diff
@@ -32,6 +32,21 @@ install(DIRECTORY include/${PROJECT_NAME}/
...
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+
+  find_package(ament_cmake_gtest REQUIRED)
+  find_package(rcpputils REQUIRED)
+  ament_add_gtest(test_get_logging_directory test/test_get_logging_directory.cpp)
+  if(TARGET test_get_logging_directory)
+    target_link_libraries(test_get_logging_directory ${PROJECT_NAME})
+    target_include_directories(test_get_logging_directory PRIVATE include)
+    #target_compile_definitions(test_get_logging_directory PUBLIC RCUTILS_ENABLE_FAULT_INJECTION)
```
ivanpauno, line 45: delete this line (?)

rcl_logging_interface/test/test_get_logging_directory.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // Fails without any relevant env vars at all (HOME included)
+  const char * directory;
+  EXPECT_EQ(RCL_LOGGING_RET_ERROR, rcl_logging_get_logging_directory(&directory, allocator));
+
+  // Default case without ROS2_LOG_DIR or ROS2_HOME being set (but with HOME)
+  rcpputils::fs::path fake_home("/fake_home_dir");
+  ASSERT_EQ(true, rcutils_set_env("HOME", fake_home.string().c_str()));
+  rcpputils::fs::path default_dir = fake_home / ".ros" / "log";
+  EXPECT_EQ(RCL_LOGGING_RET_OK, rcl_logging_get_logging_directory(&directory, allocator));
+  EXPECT_STREQ(directory, default_dir.string().c_str());
```
ivanpauno, line 68: directory is being leaked (here and later)

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ * It also expands '~' to the current user's home directory.
+ *
+ * \param[inout] directory The C string pointer at which to write the directory path.
+ *   Only meaningful if the call is successful.
+ * \param[in] allocator The allocator to use for memory allocation.
+ * \return RCL_LOGGING_RET_OK if successful, or
+ * \return RCL_LOGGING_RET_ERROR if an unspecified error occurs.
+ */
+RCL_LOGGING_INTERFACE_PUBLIC
+inline
```
ivanpauno, line 114: (feel free to ignore this comment)

I don't mind the inline implementation, in that way `rcl_logging_interface` continues being header only.
But I would rather put the implementation in a separate file, e.g.: `impl/rcl_logging_interface.inc`.
You can `#include <rcl_logging_interface/impl/rcl_logging_interface.inc` at the end of this file.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -75,36 +76,19 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-      return RCL_LOGGING_RET_ERROR;
-    }
-    if (!rcutils_mkdir(name_buffer)) {
-      RCUTILS_SET_ERROR_MSG("Failed to create user .ros directory");
+    const char * logdir;
+    rcl_logging_ret_t dir_ret = rcl_logging_get_logging_directory(&logdir, allocator);
+    if (dir_ret != RCL_LOGGING_RET_OK) {
+      // We couldn't get the log directory, so get out of here without setting up
+      // logging.
+      RCUTILS_SET_ERROR_MSG("Failed to get log directory");
```
ivanpauno, line 84: `rcl_logging_get_logging_directory` should already be setting an appropriate error message if it's not doing so.

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -132,7 +131,7 @@ rcl_logging_ret_t rcl_logging_external_initialize(
     char log_name_buffer[512] = {0};
     int print_ret = rcutils_snprintf(
       log_name_buffer, sizeof(log_name_buffer),
-      "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir, executable_name,
+      "%s/%s_%i_%" PRId64 ".log", logdir, executable_name,
```
ivanpauno, line 134: `logdir` is being leaked

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -75,36 +76,19 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
 
-    print_ret = rcutils_snprintf(name_buffer, sizeof(name_buffer), "%s/.ros/log", homedir);
-    if (print_ret < 0) {
-      RCUTILS_SET_ERROR_MSG("Failed to create log directory string");
-      return RCL_LOGGING_RET_ERROR;
-    }
-    if (!rcutils_mkdir(name_buffer)) {
-      RCUTILS_SET_ERROR_MSG("Failed to create user log directory");
+    // SPDLOG doesn't automatically create the log directories, so create them
+    rcpputils::fs::path logdir_path = std::string(logdir);
```
ivanpauno, line 89: do you need `std::string` here?

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -128,9 +112,10 @@ rcl_logging_ret_t rcl_logging_external_initialize(
       return RCL_LOGGING_RET_ERROR;
     }
 
-    print_ret = rcutils_snprintf(
+    char name_buffer[4096] = {0};
+    int print_ret = rcutils_snprintf(
       name_buffer, sizeof(name_buffer),
-      "%s/.ros/log/%s_%i_%" PRId64 ".log", homedir,
+      "%s/%s_%i_%" PRId64 ".log", logdir,
```
ivanpauno, line 118: `logdir` is being leaked

ivanpauno (commented):
I have left a bunch of comment here and there, but this seems to be going in the right direction.
Thanks for working on this @christophebedard !

ivanpauno:
By the way, the PR checker isn't passing.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +95,59 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+  if (SIZE_MAX != rcutils_find(*directory, '~')) {
+    const char * homedir = rcutils_get_home_dir();
+    if (homedir == NULL) {
+      return RCL_LOGGING_RET_ERROR;
+    }
+    *directory = rcutils_repl_str(*directory, "~", homedir, &allocator);
+    if (*directory == NULL) {
+      return RCL_LOGGING_RET_ERROR;
+    }
+  }
```
ivanpauno, line 147: (feel free to ignore)

I would rather have a function in `rcutils` for this ("expand path" or something like that).
It seems to be an useful tool on its own.

ivanpauno:
After addressing comments here, we should also update the logging tutorial: https://index.ros.org/doc/ros2/Tutorials/Logging-and-logger-configuration/.

rcl_logging_interface/CMakeLists.txt
christophebedard: done in b6013cbfb7292b11a160a770c578c00604bb70d9

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: fixed in 0705dd3a57a3e36594ac9a2d054a1b54a715fef8

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: yeah, I was wondering which way was the preferred one! Changed in 0705dd3a57a3e36594ac9a2d054a1b54a715fef8

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: fixed in 0705dd3a57a3e36594ac9a2d054a1b54a715fef8

rcl_logging_interface/CMakeLists.txt
christophebedard: sorry! Removed in ba70da858a5e9f8de3e40c40ffcd7a2d23abc3bd

christophebedard:
> By the way, the PR checker isn't passing.

yeah, it requires ros2/rcpputils#98

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: > PS: We should really have an item in the style guide clarifying that.

here? https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/

I can open an issue (and do the change too)

ivanpauno: > I can open an issue (and do the change too)

That would be great, thanks!

> yeah, I was wondering which way was the preferred one! Changed in 0705dd3a57a3e36594ac9a2d054a1b54a715fef8

The fix looks good!

christophebedard: https://github.com/ros2/ros2_documentation/issues/880

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: Sure, I can do it and come back to modify this later on (since this PR already depends on a number of other PRs :sweat_smile:): https://github.com/ros2/rcutils/issues/293

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
christophebedard: oops I forgot; fixed in cd68e6b77691137323edcc9bf473fcb35451ab32

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
christophebedard: removed (and set error messages appropriately) as part of 3d5bfb3d86138d241475940502b52179f9910628

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: done as part of 3d5bfb3d86138d241475940502b52179f9910628

I wrote quite literal error messages/doc, so let me know if that's okay.

rcl_logging_interface/test/test_get_logging_directory.cpp
christophebedard: fixed in b50d081c9d24378b47b447969fadee97da2997ba

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
christophebedard: fixed in b50d081c9d24378b47b447969fadee97da2997ba

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
christophebedard: fixed in b50d081c9d24378b47b447969fadee97da2997ba

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
wjwwood: I would prefer to avoid the inline myself... Maybe this shouldn't be in `rcl_logging_interface`? Or maybe it should just have a library...

`rmw`, by comparison, has a library `librmw` as well as providing headers which must be implemented by rmw implementations. So I think this could follow a similar pattern.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +98,73 @@ RCL_LOGGING_INTERFACE_PUBLIC
 RCUTILS_WARN_UNUSED
 rcl_logging_ret_t rcl_logging_external_set_logger_level(const char * name, int level);
 
+/// Get the logging directory.
+/**
+ * Uses various environment variables to construct a logging directory path.
+ *
+ * Use $ROS2_LOG_DIR if ROS2_LOG_DIR is set and not empty.
```
wjwwood, line 105: I don't think we should use `ROS2_` here. The environement variables we already have use just `ROS_`, e.g. `ROS_DOMAIN_ID`.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
wjwwood: Wait why are we passing by pointer?

Passing by pointer requires another check in each function that takes the pointer (check for nullptr). Since it's never ok for the allocator to be nullptr and since copying the allocator is trivial, I'd say pass by value.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +98,73 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ * Otherwise, use $ROS2_HOME/log, using ~/.ros for ROS2_HOME if not set or if empty.
+ *
+ * It also expands '~' to the current user's home directory.
+ *
+ * \param[in] allocator The allocator to use for memory allocation.
+ * \param[out] directory The C string pointer at which to write the directory path.
+ *   Only meaningful if the call is successful. Must not be null and must point to null.
```
wjwwood, line 112:
```suggestion
 *   Only meaningful if the call is successful. Must not be nullptr and must point to nullptr.
```

To avoid confusing it with `'\0'` the NUL terminating character.

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
```diff
@@ -91,6 +98,73 @@ RCL_LOGGING_INTERFACE_PUBLIC
...
+ * \return RCL_LOGGING_RET_OK if successful, or
+ * \return RCL_LOGGING_RET_INVALID_ARGUMENT if any arguments are invalid, or
+ * \return RCL_LOGGING_RET_ERROR if an unspecified error occurs.
+ */
+RCL_LOGGING_INTERFACE_PUBLIC
+inline
+rcl_logging_ret_t
+rcl_logging_get_logging_directory(const rcutils_allocator_t * allocator, const char ** directory)
+{
+  if (directory == NULL) {
```
wjwwood, line 122: nitpick: prefer the literal on the left hand side, to avoid subtle bugs like `if (directory = NULL) {`, see:

https://index.ros.org/doc/ros2/Contributing/Code-Style-Language-Versions/#style

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
wjwwood: 👍

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
ivanpauno: > Passing by pointer requires another check in each function that takes the pointer (check for nullptr). Since it's never ok for the allocator to be nullptr and since copying the allocator is trivial, I'd say pass by value.

The rationale sounds good to me, but I've seen a lot of functions in our API taking a pointer.

wjwwood: Yeah, and I've probably written several of them, but if I had to be pressed on which to prefer, I'd say by value unless you have a reason not to do so (like if the allocator is optional or there is a performance reason).

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
ivanpauno: > I would prefer to avoid the inline myself... Maybe this shouldn't be in rcl_logging_interface? 

I was thinking about adding it to another place. The only place that sounded reasonable to me was `rcl`, but in that case we cannot use the function here because it would create a circular dependecy.

> Or maybe it should just have a library...

sounds good to me

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: changed in a97d63384829af147895846a4b44346d9e48b942

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: alright, switched back to passing by value in 2a519a886c220f1619fcd239d34999ca2668c4b8

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: I added `rcutils_strdup` for `log_dir_env` and clarified the ownership of the returned pointer in 71bfafc042e5b0a742664d72a5d8a9a0565af15b

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: looking at the implementation of `rcutils_get_env`, I understand. I applied the changes to both calls to `rcutils_get_env`: 4114f12a62239519559981dafec0b6b5b7113482

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: ah yeah, I started doing that a while back, but after looking at the `rcl_logging_*` packages I assumed that it wasn't the preferred way. I'll start referring to the style guide directly from now on! Fixed in 269e9018d95a0f5f74bd1c6d642a55442f3e1670

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: good point. Switched to `ROS_*` in 18b84486e8565a7ab3a9ef319a76f52f0b1906b4 (and will do the same in my ros2/launch PR).

rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h
christophebedard: changed in 35cc88ad261cd40d97f5d3d1e4c2edba78a8d69b. I *think* I did it correctly.

christophebedard:
I think I've adressed all of your comments @ivanpauno @wjwwood!

I also added another simple test (inspired by https://github.com/ros2/launch/pull/460#discussion_r496923624) in a88529569551f9ec20ea93f46154bbb2d8350686

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,92 @@
...
+#include <rcutils/find.h>
+#include <rcutils/get_env.h>
+#include <rcutils/repl_str.h>
+#include <rcutils/strdup.h>
+
+#include "rcl_logging_interface/rcl_logging_interface.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
ivanpauno, line 27: nit: this isn't needed here.

This is only needed in pure C include files (`.h`), because they can be included from a CPP file.
In the case of a C source file, the build system picks the right compiler based on the extension.

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,92 @@
...
+    }
+    *directory = rcutils_join_path(ros_home_dir_env, "log", allocator);
+    if (NULL == *directory) {
+      RCUTILS_SET_ERROR_MSG("rcutils_join_path failed");
+      return RCL_LOGGING_RET_ERROR;
+    }
+  }
+
+  // Expand home directory
+  if (SIZE_MAX != rcutils_find(*directory, '~')) {
```
ivanpauno, line 75: should `~` be replaced only at the beginning of the string or everywhere?

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -102,12 +102,11 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-    if (homedir == nullptr) {
-      // We couldn't get the home directory; it is not really going to be
-      // possible to do logging properly, so get out of here without setting
-      // up logging.
+    char * logdir = nullptr;
+    rcl_logging_ret_t dir_ret = rcl_logging_get_logging_directory(allocator, &logdir);
+    if (RCL_LOGGING_RET_OK != dir_ret) {
+      // We couldn't get the log directory, so get out of here without setting up
+      // logging.
       return RCL_LOGGING_RET_ERROR;
```
ivanpauno, line 110:
```suggestion
      return dir_ret;
```

ivanpauno (approved):
LGTM, thanks @christophebedard !

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
christophebedard: fixed in both implementations in 47bc3e46aa7ba227a0f149f6f8871b92451d25a5

rcl_logging_interface/src/logging_dir.c
christophebedard: fixed in d0ebff761c297b824ba7244a215f42b0f97cc2e2

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,85 @@
...
+    }
+    *directory = rcutils_join_path(ros_home_dir_env, "log", allocator);
+    if (NULL == *directory) {
+      RCUTILS_SET_ERROR_MSG("rcutils_join_path failed");
+      return RCL_LOGGING_RET_ERROR;
+    }
+  }
+
+  // Expand home directory
+  if (0 == rcutils_find(*directory, '~')) {
```
ivanpauno, line 70: nitpick:
```suggestion
  if ('~' == (*directory)[0]) {
```

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,85 @@
...
+    const char * homedir = rcutils_get_home_dir();
+    if (NULL == homedir) {
+      RCUTILS_SET_ERROR_MSG("failed to get the home directory");
+      return RCL_LOGGING_RET_ERROR;
+    }
+    char * directory_not_expanded = *directory;
+    *directory = rcutils_join_path(homedir, directory_not_expanded + 2, allocator);
```
ivanpauno, line 77: what if `directory_not_expanded = "~"`?

christophebedard: what if it's `"~/"` or `"~//////"`? should I consider all those cases too?

rcl_logging_interface/src/logging_dir.c
christophebedard: changed in c2735e6bf04bd2295c0f5d03430a190a69cda7d3

rcl_logging_interface/src/logging_dir.c
christophebedard: it would be easier if there was a `rcutils` function to "clean up" a path string

ivanpauno: > what if it's "~/" or "~//////"? should I consider all those cases too?

There isn't an issue in all those cases.
But in `"~"` case, `directory_not_expanded + 2` will be accessing memory out of the buffer and likely segfault.

christophebedard: > There isn't an issue in all those cases.

yeah I realized :sweat_smile:

> But in "~" case

working on it

christophebedard: > (edit) directory_not_expanded + 2 is not a null terminated string.

even if we're assuming that `strlen(directory_not_expanded) > 2`?

ivanpauno: > even if we're assuming (and checked) that strlen(directory_not_expanded) > 2?

Not in that case, I mean only when the string is exactly "~".

christophebedard: I was trying to take care of the `"~/"` case while avoiding having two slashes, but I see how that's kinda useless too

christophebedard: alright, PTAL at 4576b1272d1f2d5bd333087915a5ca16ddc0dcd1

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,95 @@
...
+      *directory = rcutils_strdup(homedir, allocator);
+      if (NULL == *directory) {
+        RCUTILS_SET_ERROR_MSG("rcutils_strdup failed");
+        return RCL_LOGGING_RET_ERROR;
+      }
+    } else {
+      char * directory_not_expanded = *directory;
+      // Try to avoid having a double slash
+      const int offset = (2 <= strlen(*directory) && '/' == directory_not_expanded[1]) ? 2 : 1;
+      *directory = rcutils_join_path(homedir, directory_not_expanded + offset, allocator);
```
ivanpauno, line 86: Last nit, I wouldn't handle double `/`.
This snippet as-is doesn't work on Windows, and in all OS I know of double path separators are automatically simplified (e.g. `/my/home//log` works ok on Linux/macOS and `C:\my\home\\log` also works on Windows).
```suggestion
      *directory = rcutils_join_path(homedir, directory_not_expanded + 1, allocator);
```

If you want to handle this case, you would need something like https://github.com/ros2/rcutils/blob/22d5e564da8b0eae9d8c786a8e4cd78dceccb4f5/src/filesystem.c#L38-L42 (which should be moved to `filesystem.h`.

ivanpauno (approved):

rcl_logging_interface/src/logging_dir.c
christophebedard: the main reason why I did that offset thing is because it was outputing double slashes every time otherwise. I'll try to figure out why exactly, but in the meantime I've opened https://github.com/ros2/rcutils/pull/294

ivanpauno: > the main reason why I did that offset thing is because it was outputing double slashes every time otherwise

Ah, I imagine that was affecting the tests.
The issue is that the tests will fail on Windows with the current code.

You actually don't need to "join paths", you just need to join strings in this case.
Something that the following would fix the issue:
```c
size_t homedir_len = strlen(homedir);
size_t directory_not_expanded_len = strlen(directory_not_expanded);

char * new_directory = malloc(directory_not_expanded + homedir_len + 1);
// ERROR HANDLING HERE
strncpy(new_directory, homedir, homedir_len);
strncpy(new_directory + homedir_len, directory_not_expanded_len, directory_not_expanded_len + 1);
```
And you don't even need to handle the `strlen(directory_not_expanded) == 1u` case separately.
It would be better to have a join strings function in `rcutils`, but adding another blocker to this sounds unnecesary.

christophebedard: > You actually don't need to "join paths", you just need to join strings in this case.

christophebedard: > You actually don't need to "join paths", you just need to join strings in this case.

thank you for pointing it out

> It would be better to have a join strings function in rcutils

we do have `rcutils_format_string` :wink: and it can use our allocator, so I'll use that if you don't mind

christophebedard: PTAL at 93ad695580b76cddf3c510add58685e5e97e61e8

ivanpauno (approved):

christophebedard:
> `rcutils_format_string` has an implicit `2048` max length, for some strange reason (see [here](https://github.com/ros2/rcutils/blob/22d5e564da8b0eae9d8c786a8e4cd78dceccb4f5/include/rcutils/format_string.h#L32)).

I assumed 2048 was enough, but yeah there's no point in restricting it. I changed it in 49fdcd8b410e20087dcccd8eabd6853579233610

christophebedard:

christophebedard:
Hope you don't mind, here's CI for this PR + ros2/launch#460 with `--packages-above[-and-dependencies] rcl_logging_interface rcl_logging_log4cxx rcl_logging_noop rcl_logging_spdlog launch launch_xml launch_yaml`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12652/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7615/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10365/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12583/)

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,90 @@
...
+    if ('\0' == *ros_home_dir_env) {
+      ros_home_dir_env = rcutils_join_path("~", ".ros", allocator);
+      if (NULL == ros_home_dir_env) {
+        RCUTILS_SET_ERROR_MSG("rcutils_join_path failed");
+        return RCL_LOGGING_RET_ERROR;
+      }
+    }
+    *directory = rcutils_join_path(ros_home_dir_env, "log", allocator);
+    if (NULL == *directory) {
+      RCUTILS_SET_ERROR_MSG("rcutils_join_path failed");
```
clalancette, line 64: There are four memory problems here:

1.  If this second `rcutils_join_path` fails, but `ros_home_dir_env` was not set in the environment, then we leak memory.
1.  `ros_home_dir_env` is never freed if it was allocated.
1.  In the case where we are expanding the `~`, we are leaking the original memory allocated to `directory`.
1.  `ros_home_dir_env` is captured two different ways; via a memory allocation (in the case where `ROS_HOME` is not set), and via a static pointer (in the case where `ROS_HOME` is set).  This makes it impossible to do the right thing to cleanup.

All of these need to be cleaned up.  I *might* suggest splitting this function into multiple functions; I think that might make it easier to determine what should happen in each case.

clalancette (changes requested):

rcl_logging_interface/src/logging_dir.c
ivanpauno: Uff, I turned off my memory checker when reviewing this :joy:.

christophebedard: Yeah, the way it's written really doesn't make it easy. I'll go over it once again.

> 3. In the case where we are expanding the ~, we are leaking the original memory allocated to directory.

It's freed on line 83 with `directory_not_expanded` (unless I'm missing something).

christophebedard: Also maybe we should include https://github.com/ros2/rcutils/pull/298 in this PR after all :sweat_smile:

clalancette: > It's freed on line 83 with `directory_not_expanded` (unless I'm missing something).

Oh, you are totally right, sorry about that.

clalancette: > Yeah, the way it's written really doesn't make it easy. I'll go over it once again. Sorry for all the trouble.

No worries, this kind of code can be hard to get right in C.  In particular, we have to go through and be careful about the memory lifetimes here, since some of them are static and some of them are heap.  I might suggest doing something (an extra copy or something like that) to make sure they are all heap before the expansion steps.  That way it is easy to just always free it at the end.

christophebedard: fixed in 0dd91f4655e7b036337ab62826987aca55873984

I went with the "extra copy" approach (and hopefully made it more obvious); I think that should resolve all the problems you mentioned above @clalancette

christophebedard:
like @ivanpauno pointed out previously, the tests aren't passing on Windows because they're using `/` as a path separator in environment variables. I'm just wondering which solution is more appropriate:

1. change the tests to use the right path separator in the env var values, e.g. by `#ifdef`ing or calling `rcutils_to_native_path`, or
2. calling `rcutils_to_native_path` at the end of `rcl_logging_get_logging_directory`

Note that, in the *Logging and logger configuration* tutorial, I used forward slashes for Windows, so it wouldn't currently work with option 1: https://index.ros.org/doc/ros2/Tutorials/Logging-and-logger-configuration/#logging-directory-configuration

christophebedard:
I ended up going with option 2 (+ a bit of option 1, as mentioned above) because it's more resilient: c8aae9e1b9ca47f5927b0f065fbd4e12e279d364

I've also made the same modification in ros2/launch#460. Let me know what you think. I've tested on both Linux and Windows. Here's a CI rebuild:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12694/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7655/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10408/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12637/)

christophebedard:
I missed the MSBuild warning in the first CI run. Should be fixed in 1a4f0e01ab9b425c4c2db7a0cf44e792599b664d.

Windows rebuild:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12638/)

rcl_logging_interface/src/logging_dir.c
```diff
@@ -0,0 +1,106 @@
...
+      return RCL_LOGGING_RET_ERROR;
+    }
+  }
+
+  // Expand home directory
+  if ('~' == (*directory)[0]) {
+    const char * homedir = rcutils_get_home_dir();
+    if (NULL == homedir) {
+      RCUTILS_SET_ERROR_MSG("failed to get the home directory");
```
clalancette, line 81: There's a leak here on error, in that we are failing to free `directory`.

rcl_logging_interface/src/logging_dir.c
clalancette: This part is looking better, thanks.

rcl_logging_interface/test/test_get_logging_directory.cpp
```diff
@@ -0,0 +1,182 @@
...
+  explicit RestoreEnvVar(const std::string & name)
+  : name_(name),
+    value_(rcpputils::get_env_var(name.c_str()))
+  {
+  }
+
+  ~RestoreEnvVar()
+  {
+    if (!rcutils_set_env(name_.c_str(), value_.c_str())) {
+      std::cerr << "Failed to restore value of environment variable: " << name_ << std::endl;
```
clalancette, line 40: Nit: `#include <iostream>` at the top for this.

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
```diff
@@ -102,13 +102,12 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
 
-    // First get the home directory.
-    const char * homedir = rcutils_get_home_dir();
-    if (homedir == nullptr) {
-      // We couldn't get the home directory; it is not really going to be
-      // possible to do logging properly, so get out of here without setting
-      // up logging.
-      return RCL_LOGGING_RET_ERROR;
+    char * logdir = nullptr;
+    rcl_logging_ret_t dir_ret = rcl_logging_get_logging_directory(allocator, &logdir);
```
clalancette, line 106: Down below this line, if there are failures (like `rcutils_system_time_now`), we are leaking memory.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -75,36 +76,18 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-    }
-    if (!rcutils_mkdir(name_buffer)) {
-      RCUTILS_SET_ERROR_MSG("Failed to create user .ros directory");
-      return RCL_LOGGING_RET_ERROR;
+    char * logdir = nullptr;
+    rcl_logging_ret_t dir_ret = rcl_logging_get_logging_directory(allocator, &logdir);
+    if (RCL_LOGGING_RET_OK != dir_ret) {
+      // We couldn't get the log directory, so get out of here without setting up
+      // logging.
+      return dir_ret;
```
clalancette, line 84: Let's do `RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING` here.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -75,36 +76,18 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-    if (print_ret < 0) {
-      RCUTILS_SET_ERROR_MSG("Failed to create log directory string");
-      return RCL_LOGGING_RET_ERROR;
-    }
-    if (!rcutils_mkdir(name_buffer)) {
-      RCUTILS_SET_ERROR_MSG("Failed to create user log directory");
+    // SPDLOG doesn't automatically create the log directories, so create them
+    rcpputils::fs::path logdir_path(logdir);
+    if (!rcpputils::fs::create_directories(logdir_path)) {
+      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("Failed to create log directory: %s", logdir);
```
clalancette, line 90: We are leaking `logdir` on failure here.  The same goes several times below.

clalancette (changes requested):
All right, this is starting to look better.  I still found a few more memory leaks to clean up.

rcl_logging_interface/src/logging_dir.c
christophebedard: fixed as part of dd0a9768e2fd4ad15c087c849c090e4e1364509b

rcl_logging_interface/test/test_get_logging_directory.cpp
christophebedard: fixed in 49e71a001b41deeaf29a7baf6e2444c181855da7

rcl_logging_log4cxx/src/rcl_logging_log4cxx/rcl_logging_log4cxx.cpp
christophebedard: fixed as part of dd0a9768e2fd4ad15c087c849c090e4e1364509b

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
christophebedard: fixed as part of dd0a9768e2fd4ad15c087c849c090e4e1364509b

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
christophebedard: I'm guessing you mean `RCUTILS_SET_ERROR_MSG`. I had one there before but removed it when I added calls to `RCUTILS_SET_ERROR_MSG` in `rcl_logging_get_logging_directory` itself.

Anyway I did it (+ same thing for `*_log4cxx`) in 8e6661a8e8f1b7296f6b48647f3fa09b65b7dd5d

clalancette (approved):
This looks good to me now.  Thanks for the iterations.

I'll run CI on it next.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12706/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7667/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10420/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12652/)

ivanpauno:
Merging together with https://github.com/ros2/launch/pull/460, thanks @christophebedard for the contribution!

christophebedard:
thank you for the review @ivanpauno et al. :smile:

tgreier:
I would like to backport this PR to Foxy with https://github.com/ros2/launch/pull/460. However, the rcl_logging_interface was never merged into Foxy. I would need to backport quite a bit to get to this PR.
Do you think that it is ok to do that?

ivanpauno:
> I would like to backport this PR to Foxy with [ros2/launch#460](https://github.com/ros2/launch/pull/460). However, the rcl_logging_interface was never merged into Foxy. I would need to backport quite a bit to get to this PR. I would first need to backport #41 and then move forward.
> Do you think that it is ok to do that?

cc @jacobperron

tgreier:
@jacobperron @ivanpauno 
I backported (locally) https://github.com/ros2/rcl_logging/pull/41 and https://github.com/ros2/rcl_logging/pull/53 and I now get the functionality that I want from https://github.com/ros2/launch/pull/460.
Is this acceptable to do or are there additional PRs/commits that I should backport between #41 and #53?

PR #60, opened 2020-10-20
Make internal dependencies private

### LINKED ISSUE #58: spdlog dependency shouldn't need to be exported
https://github.com/ros2/rcl_logging/blob/e2456ba634223eb64aec197da7923295b468df6b/rcl_logging_spdlog/CMakeLists.txt#L64

If I understand correctly there's no reason to `ament_export_dependencies(spdlog spdlog_vendor)`. This call makes downstream packages `find_package()` them, but downstream code shouldn't need to know about them at all in order to use this logging implementation. The `<exec_depend>spdlog*` is enough to make the spdlog libraries available at runtime.

@hidmic thoughts?

sloretz:
`rcpputils`, `rcutils`, and `spdlog` don't need to be exposed to downstream users of this package at their build time because these dependencies are only used internally.

I think `rcl_logging_interface` should still be exposed because this package only appears to be useful if one has those headers, but that won't affect how this is currently used downstream because `rcl` [already depends on both](https://github.com/ros2/rcl/blob/e9c87e279bf7509c940b87fd4d6c5dff301bc735/rcl/package.xml#L19-L20) and has [custom CMake logic to `find_package()` it](https://github.com/ros2/rcl/blob/e9c87e279bf7509c940b87fd4d6c5dff301bc735/rcl/CMakeLists.txt#L18).

Fixes #58 

Uses `target_link_libraries()` because all of its dependencies are using targets already ament/ament_cmake#292

### DIFF
--- a/rcl_logging_noop/CMakeLists.txt
+++ b/rcl_logging_noop/CMakeLists.txt
@@ -25,10 +25,10 @@ set(${PROJECT_NAME}_sources
 
 add_library(${PROJECT_NAME} ${${PROJECT_NAME}_sources})
 
-ament_target_dependencies(${PROJECT_NAME}
-  rcl_logging_interface
-  rcutils
-)
+target_link_libraries(${PROJECT_NAME} PRIVATE
+  rcutils::rcutils)
+target_link_libraries(${PROJECT_NAME} PUBLIC
+  rcl_logging_interface::rcl_logging_interface)
 
 target_compile_definitions(${PROJECT_NAME} PRIVATE "RCL_LOGGING_INTERFACE_BUILDING_DLL")
 
@@ -43,7 +43,8 @@ install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}
   RUNTIME DESTINATION bin)
 
 
-ament_export_dependencies(ament_cmake rcl_logging_interface rcutils)
+# Export rcl_logging_interface to give downstream packages access to it's headers
+ament_export_dependencies(rcl_logging_interface)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(${PROJECT_NAME})
 ament_package()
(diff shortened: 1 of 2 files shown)


clalancette (commented):
The change seems reasonable to me, as long as we test `--packages-above rcl_logging_spdlog`.

I will also say that since we are doing this here, we should probably do a similar thing in `rcl_logging_log4cxx` and `rcl_logging_noop`.

hidmic (approved):
LGTM pending green CI

hidmic:
> I will also say that since we are doing this here, we should probably do a similar thing in rcl_logging_log4cxx and rcl_logging_noop.

+1

sloretz:
> I will also say that since we are doing this here, we should probably do a similar thing in rcl_logging_log4cxx and rcl_logging_noop

Added noop change in 1bce156. I didn't do `log4cxx` because it's using a find module with old-style standard CMake variables, and `ament_target_dependencies()` doesn't have a "PRIVATE" option.

sloretz:
> The change seems reasonable to me, as long as we test `--packages-above rcl_logging_spdlog`

Did you mean build packages above? If this broke something I would expect it to show up at build time. I also put `rcl` in the test args since [it has tests for logging](https://github.com/ros2/rcl/blob/master/rcl/test/rcl/test_logging.cpp).

CI (build: ` --packages-above-and-dependencies  rcl_logging_noop  rcl_logging_spdlog` test: ` --packages-select  rcl_logging_noop  rcl_logging_spdlog rcl`)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12822/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7779/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10536/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12767/)

clalancette (approved):

PR #62, opened 2020-10-29
Update QD

ahcorde:


chapulina (approved):
:top:

rcl_logging_spdlog/QUALITY_DECLARATION.md
```diff
@@ -149,15 +149,15 @@ It is **Quality Level 2**, see its [Quality Declaration document](https://github
...
+It is **Quality Level 1**, see its [Quality Declaration document](https://github.com/ros2/spdlog_vendor/blob/master/QUALITY_DECLARATION.md).
 
 ### Optional Direct Runtime ROS Dependencies [5.ii]
 
 `rcl_logging_spdlog` has no optional Direct Runtime ROS dependencies that need to be considered for this declaration.
 
 ### Direct Runtime non-ROS Dependency [5.iii]
 
-`rcl_logging_spdlog` has a Direct Runtime non-ROS dependenciy the `spdlog` library. It was declared to be Quality Level 4 [here](https://github.com/ros2/spdlog_vendor/blob/master/SPDLOG_QUALITY_DECLARATION.md).
+`rcl_logging_spdlog` has a Direct Runtime non-ROS dependenciy the `spdlog` library. It was declared to be Quality Level 1 [here](https://github.com/ros2/spdlog_vendor/blob/master/SPDLOG_QUALITY_DECLARATION.md).
```
clalancette, line 160:
```suggestion
`rcl_logging_spdlog` has a Direct Runtime non-ROS dependency on the `spdlog` library. It was declared to be Quality Level 1 [here](https://github.com/ros2/spdlog_vendor/blob/master/SPDLOG_QUALITY_DECLARATION.md).
```

clalancette (changes requested):
Besides the inline comment, we also need to:

1.  Update 1.ii, which is currently wrong.
1.  Review 3.i.  I think it doesn't have feature documentation because that is documented in `rcl_logging_interface`, but I'm not sure.

PR #51, opened 2020-09-17
Used current_path() function from rcpputils

ahcorde:
Removed TODO 

Compiling up-to `rcl_logging_spdlog` and testing `rcl_logging_spdlog` becuse this change only affect the spdlog test.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12339/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7308/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10052/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12248/)

Signed-off-by: ahcorde <ahcorde@gmail.com>

PR #69, opened 2020-12-21
[Foxy]  Allow configuring logging directory through environment variables

tgreier:
Backport to Foxy the following PRs:

- Logging interface https://github.com/ros2/rcl_logging/pull/41
- Used current_path() function from rcpputils https://github.com/ros2/rcl_logging/pull/51
- Added benchmark test to rcl_logging_spdlog https://github.com/ros2/rcl_logging/pull/52
- Allow configuring logging directory through environment variables  https://github.com/ros2/rcl_logging/pull/53


jacobperron:
This looks like a large change. It would be better to incrementally open up separate backports for each PR referenced.

Also, we should make sure that anything we backport does not break things for existing users. For example, we can't remove public headers, since we'll break any downstream code currently including them. It might be possible to workaround this issue by  including the equivalent header in `rcl_logging_interface`, if all of the function signatures are the same.

Correct me if I'm wrong, but at a glance, it looks like it should be possible to backport #53 without the other three referenced PRs. It's not as simple as cherry-picking the change, but presumably we could add equivalent API to the existing logging packages in Foxy.

tgreier:
Ok, I can kill this PR and take the approach of adding the functionality to Foxy without deleting any existing public headers (no cherry-picking).

jacobperron:
> Ok, I can kill this PR and take the approach of adding the functionality to Foxy without deleting any existing public headers (no cherry-picking).

Sounds good. I'll go ahead and close this one then.

PR #77, opened 2021-05-26
Same fix as PR #76, but for foxy specifically

nkalupahana:
Since the foxy branch is what's currently pulled when building foxy from source, this branch should also get the update from PR #76. 

(Side note, but: how do y'all typically keep all of the branches updated with the same bug fixes?)

### DIFF
--- a/rcl_logging_log4cxx/CMakeLists.txt
+++ b/rcl_logging_log4cxx/CMakeLists.txt
@@ -6,9 +6,9 @@ project(rcl_logging_log4cxx)
 if(NOT CMAKE_C_STANDARD)
   set(CMAKE_C_STANDARD 11)
 endif()
-# Default to C++14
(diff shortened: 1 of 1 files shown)


clalancette:
> (Side note, but: how do y'all typically keep all of the branches updated with the same bug fixes?)

It's on a very case-by-case basis.  Since we guarantee API *and* ABI compatibility once a distribution has been released, we have to look at the code of every patch that we want to backport.  For trivial backports, we have a bot that does it for us: https://docs.mergify.io/

But unfortunately, we can't take this one for Foxy.  Foxy explicitly only targets [C++14](https://www.ros.org/reps/rep-2000.html#foxy-fitzroy-may-2020-may-2023).  And while that says "minimum", the fact is that we can't change it at this point without having possible repercussions through the entire dependency chain.

PR #78, opened 2021-08-16
Remove rcl_logging_log4cxx.

clalancette:
We don't ever use it, and having it around causes a few problems:

1.  We have an unnecessary, and unneeded dependency on log4cxx in the
core.
2.  We can't currently build against libc++ with log4cxx.
3.  Log4cxx causes some absolute paths to be in the delivered packages
on Windows.

Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

@ivanpauno FYI, this should get rid of one more of the problems pointed out in https://github.com/ros2/rclcpp/issues/1688


clalancette:
Full CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15056/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9790/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12743/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15224/)

ivanpauno (approved):
Sounds fine to me

clalancette:
Even though the full CI was a month ago, nothing substantial has changed here.  So I'm just going to rely on that and the approval from @ivanpauno, and go ahead and merge this.  There is probably some follow-up work in the documentation and the CI jobs that needs to be done, which I'll look at next.

PR #86, opened 2022-05-08
Add rcl_logging_rcutils

sloretz:
This adds an `rcl_logging` implementation using the `rcutils` logger. This logs to stderr (by default) without touching the file system. This idea is to enable logging on read-only filesystems. The default `rcl_logging_spdlog` can't be used because it errors if it can't create a directory in `ROS_HOME`.

I thought I needed this for a personal project, but it turns out I need something a little different, so I'm making the PR in its current state. It's lacking testing and tests.


sloretz:
How do you think it should be configured? Environment variable? CMake variable?

An environment variable might be useful for not having to recompile binaries. How specific are you thinking? Very specific to the logger like `RCL_LOGGING_SPDLOG_CONSOLE_ONLY`, or very general like `ROS_READ_ONLY_FILESYSTEM` to apply to any code writing to `ROS_HOME`?

clalancette:
> An environment variable might be useful for not having to recompile binaries.

Yeah, this was my main motivation.  While rcl_logging is in some sense pluggable, in reality it is only reconfigurable at compile-time.  It would be much more useful if we could configure it at runtime instead.  With that in mind...

> How do you think it should be configured? Environment variable? CMake variable?

I think an environment variable is a good idea.

> Very specific to the logger like RCL_LOGGING_SPDLOG_CONSOLE_ONLY, or very general like ROS_READ_ONLY_FILESYSTEM to apply to any code writing to ROS_HOME?

I'm thinking it should be relatively specific, while not exposing too many internal details.  So something like `RCL_LOGGING_OUTPUT_SINK`, which currently can be `console`, `file`, `console+file`.  Then in the future if we wanted to add another target (like syslog), we have room for expansion.  What do you think about that proposal?

sloretz:
> I'm thinking it should be relatively specific, while not exposing too many internal details. So something like RCL_LOGGING_OUTPUT_SINK, which currently can be console, file, console,file. Then in the future if we wanted to add another target (like syslog), we have room for expansion. What do you think about that proposal?

Seems reasonable. It seems likely to me that only `rcl_logging_spdlog` will implement it for a while, so I'd recommend starting with an environment variable with `SPDLOG` in the name, and then making it more general if it works well.

Another general option might be to add a function to `rcl_logging_interface` that configures the output type. The advantage is it can be done in code, but a disadvantage is all the output types have to be listed in code ahead of time.

PR #102, opened 2023-08-23
Remove the last uses of ament_target_dependencies in this repo.

clalancette:
While we are here, remove unnecessary exports of
dependencies to downstream packages.  With the one exception of rcl_logging_interface, all of the packages in here are implementation-only, so they should never need to export dependencies downstream.


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13909/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20103/)

ahcorde (approved):

Issue #100, opened 2023-04-30
Is there any configuration to control max log file size?


Lannister-Xiaolin:
Same question from forum：
https://answers.ros.org/question/389402/how-to-limit-ros2-log-size/

cfveeden:
There is a PR up to expose spdlog's capabilities to control this here: https://github.com/ros2/rcl_logging/pull/103

It has been waiting for reviews for quite a while though.

Issue #92, opened 2022-08-31
proposal for how to unify logging and make it configurable with a file


wjwwood:
To start, after speaking with @clalancette, here's a bullet list of the current state of affairs:

- console messages (tuple of logger name, log level, log message, etc...) are sent to rcutils via logging macros
- rcutils determines if the logger is enabled for the console message's level
  - it checks this here: https://github.com/ros2/rcutils/blob/c49b41c9b19ae9e76e746db662285fc6c7d3b1af/src/logging.c#L911-L913
  - severity is considered here: https://github.com/ros2/rcutils/blob/c49b41c9b19ae9e76e746db662285fc6c7d3b1af/src/logging.c#L875-L889
- rcutils calls the output handler (it only supports one at a time)
  - it calls the global output handler here: https://github.com/ros2/rcutils/blob/c49b41c9b19ae9e76e746db662285fc6c7d3b1af/src/logging.c#L901-L904
- by default rcl registers a handler (so the default rcutils one is not used directly)
  - actually by default the client libraries (e.g. rclcpp and rclpy) register a single handler, but in those cases they just make it thread-safe and then forward it to the rcl handler
  - the rcl handler iterates over its internal list of output handlers and calls each one in turn with the console message
    - https://github.com/ros2/rcl/blob/191b13bc2eb8578e40799eb6b50c74b885e7066e/rcl/src/rcl/logging.c#L157-L161
  - by default it has one for the console, which just send the console message back to rcutils and it is printed there
    - https://github.com/ros2/rcl/blob/191b13bc2eb8578e40799eb6b50c74b885e7066e/rcl/src/rcl/logging.c#L91-L93
  - it also has one for rosout:
    - https://github.com/ros2/rcl/blob/191b13bc2eb8578e40799eb6b50c74b885e7066e/rcl/src/rcl/logging.c#L95-L101
  - and finally it has one for the "external logging library" which is the spdlog backend:
    - https://github.com/ros2/rcl/blob/191b13bc2eb8578e40799eb6b50c74b885e7066e/rcl/src/rcl/logging.c#L102-L113
- each logging handler does what it wants with the console message
- the spdlog backend will write it to file only

There a few problems with this:

- rcutils printing to the console means that:
  - we have to configure it separately from the logging backend (including disabling it or changing its default log level)
  - we don't take advantage of implicit features of the logging backend (like the console printing performance of spdlog)
- filter console messages in rcutils instead of in each output handler means:
[...]

PR #103, opened 2023-09-15
Spdlog rotate patch

cfveeden:
# Description
`spdlog` supports rotating loggers, but `rcl_logging` is not currently using them. This PR introduces the capability and puts the feature behind an envvar to preserve the original behaviour.

# Details
The introduced ennvars and their defaults are:
| Envvar | Default |
| --- | --- |
| RCL_LOGGING_SPDLOG_ROTATE_FILES | False |
| RCL_LOGGING_SPDLOG_ROTATING_FILE_SIZE_BYTES | 100MB |
| RCL_LOGGING_SPDLOG_MAX_NUM_FILES | 5 |

### DIFF
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -109,7 +109,7 @@ static std::string get_string_env_var(const char * env_var_name)
             std::string(
               "Failed to get env var '") + env_var_name + "': " + error_str);
   }
-  return env_var_value; // Returns empty string for unset or empty env vars
+  return env_var_value;  // Returns empty string for unset or empty env vars
 }
 
 static bool contained_in(
@@ -151,7 +151,7 @@ static bool get_bool_env_var(const char * env_var_name)
               "Unrecognized value for '") + env_var_name + "': '" + value_str + "'. " +
             "Valid truthy values: " + join_quoted(", ", "'", true_strings) + ". " +
             "Valid falsy values: " + join_quoted(", ", "'", false_strings) + ". "
-        );
+    );
   }
 }
 
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
 #include <cerrno>
 #include <chrono>
 #include <cinttypes>
@@ -23,6 +24,7 @@
 #include "rcpputils/filesystem_helper.hpp"
 #include "rcpputils/env.hpp"
 #include "rcutils/allocator.h"
+#include "rcutils/env.h"
 #include "rcutils/logging.h"
 #include "rcutils/process.h"
 #include "rcutils/snprintf.h"
@@ -30,11 +32,17 @@
 
 #include "spdlog/spdlog.h"
 #include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/sinks/rotating_file_sink.h"
 
 #include "rcl_logging_interface/rcl_logging_interface.h"
 
 static std::mutex g_logger_mutex;
 static std::shared_ptr<spdlog::logger> g_root_logger = nullptr;
+
+constexpr std::size_t DEFAULT_ROTATING_FILE_SIZE_BYTES = 100 * 1024 * 1024;
+constexpr std::size_t DEFAULT_ROTATING_MAX_NUM_FILES = 5;
+
+using std::size_t;
 
 static spdlog::level::level_enum map_external_log_level_to_library_level(int external_level)
 {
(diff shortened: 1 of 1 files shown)


cfveeden:
@clalancette and @wjwwood is this change something that looks like it can be included?

fujitatomoya (commented):
similar comment with https://github.com/ros2/rcl_logging/pull/105#pullrequestreview-1768103017

can we use `const char * config_file` argument for backend logger initialization. i think that would be the place to put those settings and configuration.

https://github.com/ros2/rcl_logging/blob/1af4e495f6faceafcd72e9c0b80e373fb901d26b/rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h#L36-L39

btw, `config_file` is not used currently, so maybe we could start, what configuration or setting should be described in `config_file` for `rcl_logging_spdlog`?

cfveeden:
@fujitatomoya `spdlog` does [not have the functionality](https://github.com/gabime/spdlog/issues/2903) to be configured from files. It seems the recommended way to configure it is either re-implement that logic, or to use `spdlog_setup`. Which is preferred for ROS2?

fujitatomoya:
@cfveeden i was thinking that we need to have something similar with https://github.com/guangie88/spdlog_setup in `rcl_logging_spdlog`. before starting implementation, maybe we can discuss on how to address this requirement in WG meeting probably [Client WG](https://docs.ros.org/en/rolling/The-ROS2-Project/Governance/Working-Groups.html#client-libraries)? what do you think?

cfveeden:
@fujitatomoya From the notes on the page the group is on hiatus and from the meeting minutes, I see the last meeting for that WG took place on 2023-01-18. Is it still active? If it is, it would be good to have the discussion and gather more requirements and insights from others.

For configuring `spdlog`, here are some pre-liminary options worth considering:
1. Re-implement the parts that make sense inside `rcl_logging_spdlog` and add features as requests come in from the ROS2 community.
2. Depend directly on `spdlog_setup`.
3. Since there are some remaining open issues on `spdlog_setup` and the [code frequency chart](https://github.com/guangie88/spdlog_setup/graphs/code-frequency) and commits shows that there hasn't been much recent activity, fork and maintain a ROS2 version of `spdlog_setup` and depend on that instead of re-implementing it.

fujitatomoya:
>  I see the last meeting for that WG took place on 2023-01-18. Is it still active?

that i did not know 😓 @gbiggs @alsora Client WG is still active? or we should have this topic to somewhere else?

alsora:
The client WG hasn't met in a few months.
I think we can discuss this on Github

fujitatomoya:
@cfveeden based on https://github.com/ros2/rcl_logging/pull/103#issuecomment-1874319169,

IMO,

> Re-implement the parts that make sense inside rcl_logging_spdlog and add features as requests come in from the ROS2 community.

this looks reasonable for me. off the top of my head,

1. what is the format configuration file, such as `yaml`?
2. what elements should be supported in configuration file?
3. deprecate the current environmental variable for spdlog, `RCL_LOGGING_SPDLOG_EXPERIMENTAL_OLD_FLUSHING_BEHAVIOR`.

maybe we can open the dedicated task issue for this and link all issues to it to track. unfortunately, i do not have bandwidth for this task right now, but i am happy to review and discuss details.

cfveeden:
@fujitatomoya

 
> 1. what is the format configuration file, such as `yaml`?
> 2. what elements should be supported in configuration file?
> 3. deprecate the current environmental variable for spdlog, `RCL_LOGGING_SPDLOG_EXPERIMENTAL_OLD_FLUSHING_BEHAVIOR`.
> 
> maybe we can open the dedicated task issue for this and link all issues to it to track. unfortunately, i do not have bandwidth for this task right now, but i am happy to review and discuss details.


1. I think either toml or yaml will be adequate. I think most people have more familiarity with yaml since it is older and has been used extensively in ROS so that's what I suggest.
2. If we are adding logging features one by one, this ticket would serve to add only the log rotation capability.
3. Should this also be a separate PR? If not how will it be deprecated?

If you want to organise the effort under an initiative, I think #92 already captures that. Especially the section under "For these reasons, I propose we make these changes:" in the OP's comment.

Other issues of interest for logging features would be #100, #105, #106.

fujitatomoya:
> I think https://github.com/ros2/rcl_logging/issues/92 provides some of the things that we should aim for, especially the section under "For these reasons, I propose we make these changes:" in the OP's comment.

agree, we can have the discussion on https://github.com/ros2/rcl_logging/issues/92

Autostone-c:
> # Description
> `spdlog` supports rotating loggers, but `rcl_logging` is not currently using them. This PR introduces the capability and puts the feature behind an envvar to preserve the original behaviour.
> 
> # Details
> The introduced ennvars and their defaults are:
> 
> Envvar	Default
> RCL_LOGGING_SPDLOG_ROTATE_FILES	False
> RCL_LOGGING_SPDLOG_ROTATING_FILE_SIZE_BYTES	100MB
> RCL_LOGGING_SPDLOG_MAX_NUM_FILES	5

hi, i follow you to this way , and i set 1*1024*1024 for 

> # Description
> `spdlog` supports rotating loggers, but `rcl_logging` is not currently using them. This PR introduces the capability and puts the feature behind an envvar to preserve the original behaviour.
> 
> # Details
> The introduced ennvars and their defaults are:
> 
> Envvar	Default
> RCL_LOGGING_SPDLOG_ROTATE_FILES	False
> RCL_LOGGING_SPDLOG_ROTATING_FILE_SIZE_BYTES	100MB
> RCL_LOGGING_SPDLOG_MAX_NUM_FILES	5

i follow you, i set ROS_LOG_DIR=~/.my_log，and i set RCL_LOGGING_SPDLOG_ROTATING_FILE_SIZE_BYTES=5*1024*1024
it does work , But there will only have some little file（70byte），by the way , my order is (ros2 launch xx  use components_container), It will result in no log files that I want to record
do you have some idea

fujitatomoya:
@Autostone-c this PR is not merged yet, and actually we are going to have discussion on how to use configuration file instead. so i think you are on your own with this PR, if you are trying.

if your problem still happens released distro, please consider to open another issue.

PR #104, opened 2023-09-25
Migrate to std::filesystem

yoneken:
https://github.com/ros2/rcpputils/issues/164
rcpputils::fs will be no longer available.
It's time to migrate to std::filesystem.

### DIFF
--- a/rcl_logging_interface/test/test_get_logging_directory.cpp
+++ b/rcl_logging_interface/test/test_get_logging_directory.cpp
@@ -14,16 +14,18 @@
 
 #include <iostream>
 #include <string>
+#include <filesystem>
 
 #include "gtest/gtest.h"
 
-#include "rcpputils/filesystem_helper.hpp"
 #include "rcpputils/env.hpp"
 #include "rcutils/allocator.h"
 #include "rcutils/env.h"
 #include "rcutils/error_handling.h"
 
 #include "rcl_logging_interface/rcl_logging_interface.h"
+
+namespace fs = std::filesystem;
 
 // This is a helper class that resets an environment
 // variable when leaving scope
@@ -79,10 +81,10 @@ TEST(test_logging_directory, directory)
   directory = nullptr;
 
   // Default case without ROS_LOG_DIR or ROS_HOME being set (but with HOME)
-  rcpputils::fs::path fake_home("/fake_home_dir");
+  fs::path fake_home("/fake_home_dir");
   ASSERT_EQ(true, rcutils_set_env("HOME", fake_home.string().c_str()));
   ASSERT_EQ(true, rcutils_set_env("USERPROFILE", fake_home.string().c_str()));
-  rcpputils::fs::path default_dir = fake_home / ".ros" / "log";
+  fs::path default_dir = fake_home / ".ros" / "log";
   EXPECT_EQ(RCL_LOGGING_RET_OK, rcl_logging_get_logging_directory(allocator, &directory));
   EXPECT_STREQ(directory, default_dir.string().c_str());
   allocator.deallocate(directory, allocator.state);
@@ -90,7 +92,7 @@ TEST(test_logging_directory, directory)
 
   // Use $ROS_LOG_DIR if it is set
   const char * my_log_dir_raw = "/my/ros_log_dir";
-  rcpputils::fs::path my_log_dir(my_log_dir_raw);
+  fs::path my_log_dir(my_log_dir_raw);
   ASSERT_EQ(true, rcutils_set_env("ROS_LOG_DIR", my_log_dir.string().c_str()));
   EXPECT_EQ(RCL_LOGGING_RET_OK, rcl_logging_get_logging_directory(allocator, &directory));
   EXPECT_STREQ(directory, my_log_dir.string().c_str());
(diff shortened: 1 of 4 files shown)


rcl_logging_interface/test/test_get_logging_directory.cpp
```diff
@@ -14,17 +14,19 @@
...
 
-#include "rcpputils/filesystem_helper.hpp"
 #include "rcpputils/env.hpp"
 #include "rcutils/allocator.h"
 #include "rcutils/env.h"
 #include "rcutils/error_handling.h"
 
 #include "rcl_logging_interface/rcl_logging_interface.h"
 
+namespace fs = std::filesystem;
```
clalancette, line 28: I'd really prefer not to do this everywhere; we use `std::` in most of the codebase, so let's just spell out `std::filesystem` everywhere.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -13,14 +13,15 @@
 // limitations under the License.
 
 #include <cerrno>
+#include <cassert>
```
clalancette, line 16: Why the addition of this header?

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -132,12 +135,17 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("Failed to create log directory: %s", logdir);
-      allocator.deallocate(logdir, allocator.state);
-      return RCL_LOGGING_RET_ERROR;
+    std::error_code ec;
+    fs::path logdir_path(logdir);
+    // std::fs::create_directory returns fail when the target directory is already exists.
+    // So, check it first.
+    if (!fs::is_directory(logdir_path, ec)) {
+      // SPDLOG doesn't automatically create the log directories, so create them
+      if (!fs::create_directories(logdir_path, ec)) {
```
clalancette, line 144: Do we need all of this?  The documentation for [create_directories](https://en.cppreference.com/w/cpp/filesystem/create_directory) says:

```
5,6) Executes (1,2) for every element of p that does not already exist. If p already exists, the function does nothing (this condition is not treated as an error).
```

clalancette (changes requested):
I like the idea in general.  I've left a few things to fix here.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
yoneken: To get an error code with a parameter without throwing an exception at L142-144.
I will change it to <system_error> for a more narrow scope.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
yoneken: IHMO Yes. 
The following code, the second output says "False".
But it is acceptable that _logdir_path_ point to a writable directory.

```
#include <filesystem>
#include <iostream>
namespace fs = std::filesystem;

int main(int argc, char **argv){
  std::cout << (fs::create_directories("./temp") ? "True" : "False") << std::endl;
  std::cout << (fs::create_directories("./temp") ? "True" : "False") << std::endl;
  return 0;
}
---
$ ./a.out 
True
False
```

yoneken: The expected path for a logging directory is **a writable directory**.

A path can have the following conditions.

1. Invalid path => unsatisfactory
2. Empty writable directory => **acceptable**
3. Empty read-only directory => unsatisfactory
4. File => unsatisfactory
5. Empty path under writable directory => **acceptable**
6. Empty path under read-only directory => unsatisfactory
7. Empty path under a file => unsatisfactory

Each return code of std::filesystem::create_directory is the following.

1. Invalid path => False
2. Empty writable directory => **False**
3. Empty read-only directory => False
4. File => False
5. Empty path under writable directory => **True**
6. Empty path under read-only directory => False
7. Empty path under a file => False

Under condition 2, we get an unexpected return code.
So, we should check the condition.


Of course, there is another way.
First: create_directory, Second: check whether the path is a directory or not.
Both ways can satisfy our expectations.

yoneken: If we passed std::filesystem::create_directories without checking the return code, test returns fail at https://github.com/yoneken/rcl_logging/blob/migrate_std/rcl_logging_spdlog/test/test_logging_interface.cpp#L95 .
Because the targetted `~/.ros` is not a writable directory (a file).

fujitatomoya (approved):
looks good to me.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -132,12 +133,17 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-    rcpputils::fs::path logdir_path(logdir);
-    if (!rcpputils::fs::create_directories(logdir_path)) {
-      RCUTILS_SET_ERROR_MSG_WITH_FORMAT_STRING("Failed to create log directory: %s", logdir);
-      allocator.deallocate(logdir, allocator.state);
-      return RCL_LOGGING_RET_ERROR;
+    std::error_code ec;
+    std::filesystem::path logdir_path(logdir);
+    // std::filesystem::create_directory returns fail when the target directory is already exists.
+    // So, check it first.
+    if (!std::filesystem::is_directory(logdir_path, ec)) {
```
sloretz, line 140: What if it's a file?

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -132,12 +133,17 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
+      // SPDLOG doesn't automatically create the log directories, so create them
+      if (!std::filesystem::create_directories(logdir_path, ec)) {
```
sloretz, line 142: https://en.cppreference.com/w/cpp/filesystem/create_directory 5 & 6 says create_director**ies** creates the directories if they don't exist otherwise it does nothing, so maybe we don't need to check if there's an existing directory at all.

clalancette (approved):
To move this forward, I switched to `create_directories` with a check of `ec`.  With that, I'm happy with this so I'm going to approve and run CI.  @yoneken I'm interested in your feedback here.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19743/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14247/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20469/)

PR #105, opened 2023-10-10
Wired in syslog into spdlog

firesurfer:
This PR wires in an additional syslog sink into the spdlog implementation. 

The syslog can be enable via the environment variable:

`RCL_LOGGING_SPDLOG_ENABLE_SYSLOG=1`

Example output in `/var/log/syslog`

```
my_pc my_node[1952055]: [WARN] [1696946734.553229518] [my_node]: my message
```

It is per default configured to log to the facility: `LOG_LOCAL1` 

By editing `/etc/rsyslog.d/50-default.conf` and adding the line

We can write the specific log into a seperate file. Alternatively it is possible to redirect the log to a remote server which is useful in enterprise environments. 
`local1.*                        -/var/log/local1.log`


firesurfer:
@wjwwood and @clalancette  sorry for pushing but I would really appreciate some feedback in order to progress :)

firesurfer:
@wjwwood  and @clalancette  sorry to be annyoing but I here am again with a friendly push

fujitatomoya (commented):
@firesurfer sorry to be late to get back to you. i got a couple of comments.
CC: @clalancette @wjwwood 

- does this work on Windows? (I do not think it does...) Windows is one of the Tier 1 platform, and spdlog backend should be working for all of them.
- i think having environmental variable for each configuration or setting is not really good idea. for that, we already have `const char * config_file` argument for backend logger initialization. i think that would be the place to put those settings and configuration.

https://github.com/ros2/rcl_logging/blob/1af4e495f6faceafcd72e9c0b80e373fb901d26b/rcl_logging_interface/include/rcl_logging_interface/rcl_logging_interface.h#L36-L39

btw, `config_file` is not used currently, so maybe we could start, what configuration or setting should be described in `config_file` for `rcl_logging_spdlog`?

cfveeden:
@fujitatomoya Does ROS2 need to support the same feature set for linux and windows? [REP2000](https://www.ros.org/reps/rep-2000.html) does not seem to say much about that. 

So if `syslog` is a linux only feature of `spdlog` then adding support for it does not seem to violate any REPs?

Can't we have adding other spdlog features like [win_eventlog_sink](https://github.com/gabime/spdlog/blob/ac55e60488032b9acde8940a5de099541c4515da/include/spdlog/sinks/win_eventlog_sink.h) as a separate PR?

Of course, checking the OS and reporting configuration errors would be required to assist users in configuring their systems properly in this case.

fujitatomoya:
> Does ROS2 need to support the same feature set for linux and windows?

IMO, basically yes but there are some platform dependent code in ROS 2 core.

> Can't we have adding other spdlog features like [win_eventlog_sink](https://github.com/gabime/spdlog/blob/ac55e60488032b9acde8940a5de099541c4515da/include/spdlog/sinks/win_eventlog_sink.h) as a separate PR?

i think we can. i think we can have `ifdef` platform statement to support syslog for linux and eventlog for windows.

firesurfer:
@fujitatomoya and @cfveeden 

So let me wrap up the results of the discussion so far:

1. We want to use the configuration file in order to configure such features instead of using environment variables
2. I `ifdef` the syslog support to be only included on linux -> In a seperate PR we can add eventlog for windows


Regarding the configuration file:

At the moment there is only support for passing in the path to a configuration file. There is no definition in which format the configuration file should be. 

The documentation actually says:
> Note that the format of this file is backend-specific (and is currently unimplemented for the default backend logger of spdlog)

What kind of format would you prefer ? I would suggest using yaml file as they are widely used all over the ROS2 ecosystem. On the other hand this would mean to introduce a new dependency to `yaml-cpp`. 

How should existing parameters be handled ?  I already stated my suggestion above:

> My guess would be that only adding new parameters to the config file makes sense in order to not break compatibility. As an alternative we could define for currently implemented parameters that the environment variables will always overwrite their counter pendants in the config file.

If I can find some time for that I will try to wire in the config file support in a separate PR.

cfveeden:
@firesurfer. I want to make you aware of a similar discussion going on in #103. Can we perhaps coordinate the various additions under #92?

firesurfer:
@cfveeden If I understand the discussion in #92 correctly it is unclear when a common solution for the log configuration will be found / worked on. Does this mean no addition to the rcl_logging system can be made until at some point in the future this solution has been found? (Sorry for being a bit provocative there) 

The reason I am pushing towards having a syslog functionality is that it is hard at the moment to really get a complete log of an application (without starting to manually collect multiple logfiles from multiple PCs)  which can be necessary for debugging certain issues.

130s (changes requested):
I like the motive, thanks for putting this together. Added a few potentially simple coding questions.

For Windows, there was a recent change announced [discourse.ros.org#39502](https://discourse.ros.org/t/windows-survey-results/39502) but I'm not sure how/if that affects this sort of code change.

One question: Like I described one example of logging pipeline [answers.ros.org#371667](https://answers.ros.org/question/371667/ros-and-syslog/) (of ros1, but I hope the flow of pipeline is the same in ros2), there's a way to configure in OS-system level outside of ROS, in order for logs to be flushed to `syslog`. In my experience in product dev that sort of flexibility was important.

With this change, if ROS is configured to log onto `syslog`, would that affect system level configuration for logging? E.g. If my ROS2 application is running as `systemd` service AND ROS is configured to log to syslog, would I see logs duplicated in `syslog`?

firesurfer:
> For Windows, there was a recent change announced [discourse.ros.org#39502](https://discourse.ros.org/t/windows-survey-results/39502) but I'm not sure how/if that affects this sort of code change.

The discussion points in this thread are still valid. In some way either it should produce an error on windows or fallback to the windows eventlog


> With this change, if ROS is configured to log onto syslog, would that affect system level configuration for logging? E.g. If my ROS2 application is running as systemd service AND ROS is configured to log to syslog, would I see logs duplicated in syslog?

I can give you a clear yes and no answer.  It depends on the configuration of your local syslog daemon. 
If you log from multiple sources to the default syslog then the answer is yes. If you log to a different sink then the answer is probably not. You could also disable console logging for the ROS application in that case. 


Regarding the code suggestions -> I think it does not make sense to discuss any of the code until https://github.com/ros2/rcl_logging/issues/92 came to a conclusion.

PR #106, opened 2023-10-25
Add RCL_LOGGING_DISABLE_FILE_WRITING to disable file logging

mganeff:
Add ability to completely disable file writing in rcp_logging_spdlog by setting the environment variable:

`RCL_LOGGING_DISABLE_FILE_WRITING=1`

### DIFF
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -95,6 +95,14 @@ rcl_logging_ret_t rcl_logging_external_initialize(
   const char * config_file,
   rcutils_allocator_t allocator)
 {
+  try {
+    if ("1" == rcpputils::get_env_var("RCL_LOGGING_DISABLE_FILE_WRITING")) {
+      // explicitly true
+      return RCL_LOGGING_RET_OK;
+    }
+  } catch (const std::runtime_error & error) {
+  }
+
   std::lock_guard<std::mutex> lk(g_logger_mutex);
   // It is possible for this to get called more than once in a process (some of
   // the tests do this implicitly by calling rclcpp::init more than once).
@@ -196,22 +204,27 @@ rcl_logging_ret_t rcl_logging_external_initialize(
 
 rcl_logging_ret_t rcl_logging_external_shutdown()
 {
-  spdlog::drop("root");
-  g_root_logger = nullptr;
+  if (g_root_logger != nullptr) {
+    spdlog::drop("root");
+    g_root_logger = nullptr;
+  }
   return RCL_LOGGING_RET_OK;
 }
 
 void rcl_logging_external_log(int severity, const char * name, const char * msg)
 {
   (void)name;
-  g_root_logger->log(map_external_log_level_to_library_level(severity), msg);
+  if (g_root_logger != nullptr) {
+    g_root_logger->log(map_external_log_level_to_library_level(severity), msg);
+  }
 }
 
 rcl_logging_ret_t rcl_logging_external_set_logger_level(const char * name, int level)
 {
   (void)name;
-
-  g_root_logger->set_level(map_external_log_level_to_library_level(level));
+  if (g_root_logger != nullptr) {
+    g_root_logger->set_level(map_external_log_level_to_library_level(level));
+  }
 
   return RCL_LOGGING_RET_OK;
 }


clalancette:
So I have to say that this is a weird feature.

The entire purpose of the `rcl_logging` layer is to write to disk.  Writing to the console and `/rosout` are done at different layers.  And we already have a way to disable writing to disk at the RCL layer: http://docs.ros.org/en/rolling/Concepts/Intermediate/About-Logging.html#node-creation

So with all of that in mind, can you please explain the use-case for this?

mganeff:
I understand that I can disable logging for a certain node by using the **disable-external-lib-logs** argument

`ros2 run foo bar --ros-args --disable-external-lib-logs`

I am however unable to find a working solution for disabling logging for nodes started with _ros2 launch_

- ros2 launch launch_file.launch (xml based)
- ros2 launch launch_file.launch.py

In our environment we spawn multiple nodes (20+). For testing we would like to create the logs but for deployment we would like to disable them, hence the env variable to be able to disable/enable external logging features for all nodes :)

clalancette:
> I am however unable to find a working solution for disabling logging for nodes started with _ros2 launch_

You should be able to declare a launch argument for that, and pass it into all of the nodes.  You can see some documentation about that in http://docs.ros.org/en/rolling/Tutorials/Intermediate/Launch/Using-ROS2-Launch-For-Large-Projects.html or http://docs.ros.org/en/rolling/Tutorials/Intermediate/Launch/Using-Substitutions.html

mganeff:
Thanks for the answer. I managed to disable logging for all my nodes by adjusting the launch files:

ros2 launch .xml:
```
<node pkg="foo" exec="bar" output="screen" namespace="hub" ros_args="--disable-external-lib-logs"/>
```
ros2 launch .py:
 ```
 ld.add_action(
    Node(
        package="foo",
        executable="bar",
        output="screen",
        emulate_tty=True,
        arguments=["--ros-args", "--disable-external-lib-logs"],
    )
)
```
ros2 run:
```
ros2 run foo bar --ros-args --disable-external-lib-logs
```
I was wondering if there is an environment variable such like **DEFAULT_ROS_ARGS** which I could extend with _--disable-external-lib-logs_ to be able to toggle logging on/off for our nodes.

mganeff:
Despite the parameter disable-external-lib-logs, still logs are generated in ~/.ros/log. Each launched node gets a directory with a launch.log. Is there a way to disable this?

clalancette:
> Despite the parameter disable-external-lib-logs, still logs are generated in ~/.ros/log. Each launched node gets a directory with a launch.log. Is there a way to disable this?

That's the logging coming from `launch`, rather than from the node itself.  I think if you remove `output` from your launch statements, that will stop happening.

fujitatomoya (commented):
i think even if we had a feature something like this, doing this in the specific implementation is not the right place to do that.

> I was wondering if there is an environment variable such like DEFAULT_ROS_ARGS which I could extend with --disable-external-lib-logs to be able to toggle logging on/off for our nodes.

no i don't think so. for me this can be scary giant switch if we have it. because we can add remapping, namespace, names, parameters and logging configuration via arguments. i think those arguments are meant to be applied to the `Node`, not any `Node`. So this could possibly lead to the problem for user.

ciandonovan:
> > Despite the parameter disable-external-lib-logs, still logs are generated in ~/.ros/log. Each launched node gets a directory with a launch.log. Is there a way to disable this?
> 
> That's the logging coming from `launch`, rather than from the node itself. I think if you remove `output` from your launch statements, that will stop happening.

I'm very surprised there isn't just a simple way to turn off all disk writing in ROS 2 globally with something like an env var—what are users on read-only embedded systems supposed to do? Or those with small disks who are worried about storage exhaustion?

https://robotics.stackexchange.com/questions/104972/how-can-i-disable-all-logging-to-files-under-ros-but-still-log-to-stderr

AFAICT this is the only guaranteed solution, and it's a complete hack. You shouldn't need a custom third-party kernel module just to ensure ROS 2 doesn't write anything to disk!


> So I have to say that this is a weird feature.
> 
> The entire purpose of the `rcl_logging` layer is to write to disk. Writing to the console and `/rosout` are done at different layers. And we already have a way to disable writing to disk at the RCL layer: http://docs.ros.org/en/rolling/Concepts/Intermediate/About-Logging.html#node-creation
> 
> So with all of that in mind, can you please explain the use-case for this?

It's not a great solution as-is, having to ensure that that ROS 2 argument is added to every single node on the system, rather than just turning it off globally with an env var. What if I'm using upstream node/launch files unmodified? As a sysadmin I can be guaranteed with an env var that nothing will be written, instead here if a dev slips through a node without explicitly turning off disk logging accidentally, I'll have a hung device in the field in a few weeks/months time.

Is RCL the right place to submit a PR to optionally completely disable any disk writes?

clalancette:
> Is RCL the right place to submit a PR to optionally completely disable any disk writes?

But we already have this ability with the command-line switch.  Are you advocating for adding an environment variable that will accomplish the same thing?

ciandonovan:
> > Is RCL the right place to submit a PR to optionally completely disable any disk writes?
> 
> But we already have this ability with the command-line switch. Are you advocating for adding an environment variable that will accomplish the same thing?

Yes, because that can more readily be applied across a system, that might contain hundreds of nodes. Also it is more amenable to containers with `--env` overwrite flags.

In general I'd like to see an env var that can guarantee no writes are made from any part of the stack, as even with `--disable-external-lib-logs` ROS 2 Launch still writes files, as might other subsystems.

To protect from disk exhaustion and for read-only embedded systems, any file write attempt anywhere in ROS 2 should really read first this proposed env var before continuing or not—a global kill switch.

PR #109, opened 2023-12-13
add file_name_prefix parameter to external log configuration.

fujitatomoya:
address https://github.com/ros2/ros2cli/issues/856

### DIFF
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -29,6 +29,7 @@
 #include "rcutils/logging.h"
 #include "rcutils/process.h"
 #include "rcutils/snprintf.h"
+#include "rcutils/strdup.h"
 #include "rcutils/time.h"
 
 #include "spdlog/spdlog.h"
@@ -167,24 +168,25 @@ rcl_logging_ret_t rcl_logging_external_initialize(
     bool file_name_provided = (nullptr != file_name_prefix) && (file_name_prefix[0] != '\0');
     char * basec;
     if (file_name_provided) {
-      basec = const_cast<char *>(file_name_prefix);
+      basec = rcutils_strdup(file_name_prefix, allocator);
     } else {  // otherwise, get the program name.
       basec = rcutils_get_executable_name(allocator);
-      if (basec == nullptr) {
-        // We couldn't get the program name, so get out of here without setting up
-        // logging.
-        RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
-        return RCL_LOGGING_RET_ERROR;
-      }
     }
+    if (basec == nullptr) {
+      // We couldn't get the program name, so get out of here without setting up
+      // logging.
+      RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
+      return RCL_LOGGING_RET_ERROR;
+    }
+    RCPPUTILS_SCOPE_EXIT(
+    {
+      allocator.deallocate(basec, allocator.state);
+    });
     char name_buffer[4096] = {0};
     int print_ret = rcutils_snprintf(
       name_buffer, sizeof(name_buffer),
       "%s/%s_%i_%" PRId64 ".log", logdir,
       basec, rcutils_get_pid(), ms_since_epoch);
-    if (!file_name_provided) {
-      allocator.deallocate(basec, allocator.state);
-    }
     if (print_ret < 0) {
       RCUTILS_SET_ERROR_MSG("Failed to create log file name string");
       return RCL_LOGGING_RET_ERROR;
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -182,12 +182,12 @@ rcl_logging_ret_t rcl_logging_external_initialize(
       name_buffer, sizeof(name_buffer),
       "%s/%s_%i_%" PRId64 ".log", logdir,
       basec, rcutils_get_pid(), ms_since_epoch);
+    if (!file_name_provided) {
+      allocator.deallocate(basec, allocator.state);
+    }
     if (print_ret < 0) {
       RCUTILS_SET_ERROR_MSG("Failed to create log file name string");
       return RCL_LOGGING_RET_ERROR;
-    }
-    if (!file_name_provided) {
-      allocator.deallocate(basec, allocator.state);
     }
 
     auto sink = std::make_unique<spdlog::sinks::basic_file_sink_mt>(name_buffer, false);
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -95,6 +95,7 @@ get_should_use_old_flushing_behavior()
 }  // namespace
 
 rcl_logging_ret_t rcl_logging_external_initialize(
+  const char * file_name_prefix,
   const char * config_file,
   rcutils_allocator_t allocator)
 {
@@ -163,19 +164,19 @@ rcl_logging_ret_t rcl_logging_external_initialize(
     }
     int64_t ms_since_epoch = RCUTILS_NS_TO_MS(now);
 
-    // Get the program name.
-    char * basec = rcutils_get_executable_name(allocator);
-    if (basec == nullptr) {
-      // We couldn't get the program name, so get out of here without setting up
-      // logging.
-      RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
-      return RCL_LOGGING_RET_ERROR;
+    bool file_name_provided = (nullptr != file_name_prefix) && (file_name_prefix[0] != '\0');
+    char * basec;
+    if (file_name_provided) {
+      basec = const_cast<char *>(file_name_prefix);
+    } else {  // otherwise, get the program name.
+      basec = rcutils_get_executable_name(allocator);
+      if (basec == nullptr) {
+        // We couldn't get the program name, so get out of here without setting up
+        // logging.
+        RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
+        return RCL_LOGGING_RET_ERROR;
+      }
     }
-    RCPPUTILS_SCOPE_EXIT(
-    {
-      allocator.deallocate(basec, allocator.state);
-    });
-
     char name_buffer[4096] = {0};
     int print_ret = rcutils_snprintf(
       name_buffer, sizeof(name_buffer),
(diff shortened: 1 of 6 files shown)


rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -163,19 +164,19 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-      // We couldn't get the program name, so get out of here without setting up
-      // logging.
-      RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
-      return RCL_LOGGING_RET_ERROR;
+    bool file_name_provided = (nullptr != file_name_prefix) && (file_name_prefix[0] != '\0');
+    char * basec;
+    if (file_name_provided) {
+      basec = const_cast<char *>(file_name_prefix);
+    } else {  // otherwise, get the program name.
+      basec = rcutils_get_executable_name(allocator);
```
fujitatomoya, line 172: falls back to current behavior, so everything should work as before. but this is API break change, needs to rebuild.

rcl_logging_spdlog/test/test_logging_interface.cpp
```diff
@@ -111,7 +117,7 @@ TEST_F(LoggingTest, init_old_flushing_behavior)
   RestoreEnvVar env_var("RCL_LOGGING_SPDLOG_EXPERIMENTAL_OLD_FLUSHING_BEHAVIOR");
   rcpputils::set_env_var("RCL_LOGGING_SPDLOG_EXPERIMENTAL_OLD_FLUSHING_BEHAVIOR", "1");
 
-  ASSERT_EQ(RCL_LOGGING_RET_OK, rcl_logging_external_initialize(nullptr, allocator));
+  ASSERT_EQ(RCL_LOGGING_RET_OK, rcl_logging_external_initialize(nullptr, nullptr, allocator));
```
fujitatomoya, line 120: I think we should add more test cases with specifying the file name prefix argument, and check that file does exist in the log directory.

fujitatomoya:
@iuhilnehc-ynos requesting review.

iuhilnehc-ynos (commented):
a minior issue, otherwise LGTM.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -185,6 +186,9 @@ rcl_logging_ret_t rcl_logging_external_initialize(
       RCUTILS_SET_ERROR_MSG("Failed to create log file name string");
       return RCL_LOGGING_RET_ERROR;
     }
+    if (!file_name_provided) {
+      allocator.deallocate(basec, allocator.state);
+    }
```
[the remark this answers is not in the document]
fujitatomoya: https://github.com/ros2/rcl_logging/pull/109/commits/6e1a531fef5b4e5f6cffe6c67219fcb9b16567ab fixes it.

i did not use `RCPPUTILS_SCOPE_EXIT ` here, just calling `deallocate`.

rcl_logging_spdlog/test/fixtures.hpp
```diff
@@ -35,7 +35,7 @@
 #define pclose _pclose
 #define DIR_CMD "dir /B"
 #else
-#define DIR_CMD "ls -d"
+#define DIR_CMD "ls -td"
```
fujitatomoya, line 38: actually this was a bug, the original tests were NOT comparing the latest files against the expected result. since this will list the files in time ascending order, i happened to find this when adding the test. this will align the same behavior with windows, to list time descending order.

fujitatomoya:
@iuhilnehc-ynos requesting 2nd review.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -182,13 +182,13 @@ rcl_logging_ret_t rcl_logging_external_initialize(
       name_buffer, sizeof(name_buffer),
       "%s/%s_%i_%" PRId64 ".log", logdir,
       basec, rcutils_get_pid(), ms_since_epoch);
+    if (!file_name_provided) {
```
iuhilnehc-ynos, line 185: It can work, but it seems not a good practice as what if there is some extra logic that can return between `basec` allocation and `rcutils_snprintf` in the future.

iuhilnehc-ynos (approved):

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
fujitatomoya: yeah i thought that too.

the things are,
- we cannot use if statement with `RCPPUTILS_SCOPE_EXIT`, since it will deallocate before used.
- besides, i do want to avoid unnecessary memory allocation when we can use `file_name_prefix`.

there are some other places to do deallocation like this, and it is really close not to miss in this case.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19988/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14454/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20697/)

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
iuhilnehc-ynos:
```cpp
    RCPPUTILS_SCOPE_EXIT(
    {
      if (!file_name_provided) {
        allocator.deallocate(basec, allocator.state);
      }
    });
```

> we cannot use if statement with RCPPUTILS_SCOPE_EXIT, since it will deallocate before used.

I am not sure which way you used, is it like below,

```cpp
    if (!file_name_provided) {
      RCPPUTILS_SCOPE_EXIT(
      {
        allocator.deallocate(basec, allocator.state);
      });
    }
```

> there are some other places to do deallocation like this.

They're not a good practice as well.

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -163,24 +164,27 @@ rcl_logging_ret_t rcl_logging_external_initialize(
...
-    char * basec = rcutils_get_executable_name(allocator);
-    if (basec == nullptr) {
-      // We couldn't get the program name, so get out of here without setting up
-      // logging.
-      RCUTILS_SET_ERROR_MSG("Failed to get the executable name");
-      return RCL_LOGGING_RET_ERROR;
+    bool file_name_provided = (nullptr != file_name_prefix) && (file_name_prefix[0] != '\0');
+    char * basec;
+    if (file_name_provided) {
+      basec = const_cast<char *>(file_name_prefix);
```
clalancette, line 170: I think this would be much better as:

```suggestion
      basec = rcutils_strdup(file_name_prefix, allocator);
```

That way, we can leave the `RCPPUTILS_SCOPE_EXIT` in place below and it will all work properly.  Yes, this is an unnecessary copy, but since this is only called once-per-context, that shouldn't really matter.

rcl_logging_spdlog/test/fixtures.hpp
```diff
@@ -109,15 +109,22 @@ class LoggingTest : public AllocatorTest
...
-    if (nullptr == exe_name) {
-      throw std::runtime_error("Failed to determine executable name");
+    char * exe_name;
+    if (name == nullptr || name[0] == '\0') {
+      exe_name = rcutils_get_executable_name(allocator);
+      if (nullptr == exe_name) {
+        throw std::runtime_error("Failed to determine executable name");
+      }
+    } else {
+      exe_name = const_cast<char *>(name);
```
clalancette, line 121: Same here; it is just easier to do the duplication, and then unconditionally free later on.

clalancette (changes requested):
Besides what I left inline, Windows CI doesn't seem too happy with this change.  We'll also need to fix that.

rcl_logging_spdlog/test/fixtures.hpp
fujitatomoya: addressed with https://github.com/ros2/rcl_logging/pull/109/commits/1c6dffc12b6d9342b04cbd072fcf54aab63f918f

rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
fujitatomoya: addressed with https://github.com/ros2/rcl_logging/pull/109/commits/1c6dffc12b6d9342b04cbd072fcf54aab63f918f

rcl_logging_spdlog/test/fixtures.hpp
```diff
@@ -28,14 +28,15 @@
...
 #include "rcutils/error_handling.h"
 #include "rcutils/process.h"
+#include "rcutils/strdup.h"
 #include "rcutils/types/string_array.h"
 
 #ifdef _WIN32
 #define popen _popen
 #define pclose _pclose
-#define DIR_CMD "dir /B"
+#define DIR_CMD "dir /B /O-D"
```
fujitatomoya, line 37: windows needs some help as well, my assumption was incorrect... 😓 

```batch
C:\Users\mrx\work>type nul > 1000

C:\Users\mrx\work>type nul > 1001

C:\Users\mrx\work>dir /B
1000
1001

C:\Users\mrx\work>dir /B /O-D
1001
1000
```

here we need to get the latest log file to compare against the expected strings, adding `/O-D` can make sure listing time descending order.

fujitatomoya:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20005/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14469/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20713/)

fujitatomoya:
@clalancette requesting review again, windows failure has been addressed and CI is green.

this needs to go with https://github.com/ros2/rcl/pull/1127

clalancette (approved):

ZhenshengLee:
Happy to see the feature, I wonder if this feature will be applied to LTS Humble?

fujitatomoya:
i do not think we can do that, this breaks API.

ZhenshengLee:
> i do not think we can do that, this breaks API.

Thanks for your reply! Which official version will this feature be applied/released to? Iron or the future J version?

fujitatomoya:
[Jazzy Jalisco](https://docs.ros.org/en/rolling/Releases/Release-Jazzy-Jalisco.html), see distributions for https://docs.ros.org/en/rolling/Releases.html#future-distributions.

cz000:
> i do not think we can do that, this breaks API.

How can I fix this issue in LTS Humble? Do you have any idea?

fujitatomoya:
@cz000 this change breaks ABI compatibility, so we cannot backport this fix to humble or iron. something you can do is to cherry-pick the fix and build the package from the humble source code.

PR #110, opened 2024-02-28
Improve error message handling for spdlog

christophebedard:
Kind of relates to https://github.com/ros2/rcutils/issues/269

1. Modify tests to _expect_ the error message to be set when it should be set and then reset it
2. In `rcl_logging_spdlog`'s `rcl_logging_external_initialize()`: in case of `rcl_logging_get_logging_directory()` failure, append error message instead of overwriting it without resetting it (and thus generating a "This error state is being overwritten" warning)

This results in `rcl_logging_spdlog/test_test_logging_interface.cpp` not generating any "This error state is being overwritten" warnings.


christophebedard:
This is a draft pending higher-level discussion for overall error message handling.

clalancette (commented):
In short, this seems reasonable to me.  I'm going to hold off on approving since this is still a draft and we need to have the higher-level discussion.

fujitatomoya (commented):
draft looks good to me

PR #111, opened 2024-02-29
Re-order rcl_logging_interface include

christophebedard:
We're not very strict about includes, but I noticed a slight divergence as a result of https://github.com/ros2/rcl_logging/pull/41


clalancette (approved):
Sure (I don't have a strong opinion here).

clalancette:
Merging just with the Rpr job passing.

PR #116, opened 2024-03-06
Check allocator validity in some rcl_logging functions

cottsay:
If the allocator is zero-initialized, it may cause a segfault when it is used later in the function. Right now, we're relying on other `rcutils_*` functions which use the allocator to fail before we attempt to use it directly, but just because the allocator is passed to these functions doesn't always mean that it is used, especially when the program flow can change based on things like environment variables.

### DIFF
--- a/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
+++ b/rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
@@ -100,6 +100,8 @@ rcl_logging_ret_t rcl_logging_external_initialize(
   const char * config_file,
   rcutils_allocator_t allocator)
 {
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RCL_LOGGING_RET_INVALID_ARGUMENT);
+
   std::lock_guard<std::mutex> lk(g_logger_mutex);
   // It is possible for this to get called more than once in a process (some of
   // the tests do this implicitly by calling rclcpp::init more than once).
--- a/rcl_logging_interface/src/logging_dir.c
+++ b/rcl_logging_interface/src/logging_dir.c
@@ -28,6 +28,7 @@ rcl_logging_get_logging_directory(rcutils_allocator_t allocator, char ** directo
     RCUTILS_SET_ERROR_MSG("directory argument must not be null");
     return RCL_LOGGING_RET_INVALID_ARGUMENT;
   }
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RCL_LOGGING_RET_INVALID_ARGUMENT);
   if (NULL != *directory) {
     RCUTILS_SET_ERROR_MSG("directory argument must point to null");
     return RCL_LOGGING_RET_INVALID_ARGUMENT;
--- a/rcl_logging_spdlog/test/test_logging_interface.cpp
+++ b/rcl_logging_spdlog/test/test_logging_interface.cpp
@@ -74,7 +74,7 @@ TEST_F(LoggingTest, init_invalid)
     rcl_logging_external_initialize("anything", nullptr, bad_allocator));
   rcutils_reset_error();
   EXPECT_EQ(
-    RCL_LOGGING_RET_ERROR,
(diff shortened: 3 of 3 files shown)


rcl_logging_spdlog/src/rcl_logging_spdlog.cpp
```diff
@@ -100,6 +100,8 @@ rcl_logging_ret_t rcl_logging_external_initialize(
   const char * config_file,
   rcutils_allocator_t allocator)
 {
+  RCUTILS_CHECK_ALLOCATOR(&allocator, return RCL_LOGGING_RET_ERROR);
```
cottsay, line 103: Note that an "invalid argument" return code would be more appropriate, but this error code has already been established for at least some cases of an invalid allocator.

fujitatomoya: the same check can be applied to `rcl_logging_get_logging_directory` as a function scope?

clalancette: > Note that an "invalid argument" return code would be more appropriate, but this error code has already been established for at least some cases of an invalid allocator.

Doing a quick grep for `RCUTILS_CHECK_ALLOCATOR` through our codebase, it looks like we most often use `INVALID_ARGUMENT` as the return value.  So I'd suggest that we switch to that here, as well as update the documentation for `rcl_logging_external_initialize` to say that it is a possible return code.

> the same check can be applied to `rcl_logging_get_logging_directory` as a function scope?

Yeah, I think that is a good idea.

cottsay: Just so we're clear: this will _technically_ be an API break as is demonstrated by the need to update the test.

clalancette: > Just so we're clear: this will _technically_ be an API break as is demonstrated by the need to update the test.

Yeah, I think that is OK here.  This is a really low-level function, and I don't really expect anything outside of `rcl` to be calling it.  Indeed, looking through all of the packages released into Rolling, nothing else calls it.

I also looked through some external rcl* packages:

* [ros2-rust](https://github.com/search?q=repo%3Aros2-rust%2Fros2_rust%20rcl_logging_external_initialize&type=code) - doesn't use it
* [ros2-java](https://github.com/search?q=repo%3Aros2-java%2Fros2_java%20rcl_logging_external_initialize&type=code) - doesn't use it
* [ada-ros](https://github.com/search?q=repo%3Aada-ros%2Frclada%20rcl_logging_external_initialize&type=code) - does use it, but since the repo hasn't been updated in 3 years I think there will be other breakage as well
* [r2r](https://github.com/search?q=repo%3Asequenceplanner%2Fr2r%20rcl_logging_external_initialize&type=code) - doesn't use it
* [rclc](https://github.com/search?q=repo%3Aros2%2Frclc%20rcl_logging_external_initialize&type=code) - doesn't use it
* [rclnodejs](https://github.com/search?q=repo%3ARobotWebTools%2Frclnodejs%20rcl_logging_external_initialize&type=code) - doesn't use it

I also did a [full search on GitHub](https://github.com/search?q=rcl_logging_external_initialize&type=code) for it, and while there are a few more uses of it, I don't think anything catastrophic is going to happen if we add this return code here.

fujitatomoya (approved):
lgtm

cottsay:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/20431/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/14833/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/21174/)

clalancette (approved):

clalancette:
The one thing I'm going to suggest here is that we run tests `--packages-above rcl_logging_interface rcl_logging_spdlog`.  While my look above suggested that we won't have a problem, I think we should double check via CI before merging.

Issue #121, opened 2024-09-23
Issue with the generated CMake config files


gergondet-woven:
Hello,

With https://github.com/ros2/rcl_logging/pull/102, the dependency of `rcl_logging_spdlog` upon `rcl_logging_interface` was removed for downstream users. The reasoning behind the PR makes sense but the library still has one public dependency on `rcl_logging_interface`, defined [here](https://github.com/ros2/rcl_logging/blob/rolling/rcl_logging_spdlog/CMakeLists.txt#L37).

This means that when one calls `find_package(rcl_logging_spdlog REQUIRED)` in an isolated setting*, the `rcl_logging_spdlog` library links to an unknown library.

I think this can be simply addressed by limiting the link to the build interface, i.e.:
```cmake
target_link_libraries(${PROJECT_NAME} PUBLIC
  $<BUILD_INTERFACE:rcl_logging_interface::rcl_logging_interface>)
```
or by making this link private and linking the test with `rcl_logging_interface` explicitly.

*note: I realize this not a very common scenario, we came across this issue because we use [these Bazel rules](https://github.com/RobotLocomotion/drake-ros/tree/main/bazel_ros2_rules) which uses CMake in this way to scrape underlying colcon workspace.

Thanks in advance for taking this into consideration and have a good day!

fujitatomoya:
@gergondet-woven https://github.com/ros2/rcl_logging/pull/122 is created, it would be appreciated if you give me a feedback.

PR #122, opened 2024-09-23
rcl_logging_interface is only valid path with build environment.

fujitatomoya:
closes https://github.com/ros2/rcl_logging/issues/121

### DIFF
--- a/rcl_logging_noop/CMakeLists.txt
+++ b/rcl_logging_noop/CMakeLists.txt
@@ -25,7 +25,7 @@ add_library(${PROJECT_NAME} src/rcl_logging_noop.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE
   rcutils::rcutils)
 target_link_libraries(${PROJECT_NAME} PUBLIC
-  $<BUILD_INTERFACE:rcl_logging_interface::rcl_logging_interface>)
+  rcl_logging_interface::rcl_logging_interface)
 
 target_compile_definitions(${PROJECT_NAME} PRIVATE "RCL_LOGGING_INTERFACE_BUILDING_DLL")
 
@@ -39,7 +39,7 @@ install(TARGETS ${PROJECT_NAME} EXPORT ${PROJECT_NAME}
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin)
 
-
+ament_export_dependencies(rcl_logging_interface)
 ament_export_libraries(${PROJECT_NAME})
 ament_export_targets(${PROJECT_NAME})
 ament_package()
--- a/rcl_logging_noop/CMakeLists.txt
+++ b/rcl_logging_noop/CMakeLists.txt
@@ -25,7 +25,7 @@ add_library(${PROJECT_NAME} src/rcl_logging_noop.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE
   rcutils::rcutils)
 target_link_libraries(${PROJECT_NAME} PUBLIC
-  rcl_logging_interface::rcl_logging_interface)
+  $<BUILD_INTERFACE:rcl_logging_interface::rcl_logging_interface>)
 
 target_compile_definitions(${PROJECT_NAME} PRIVATE "RCL_LOGGING_INTERFACE_BUILDING_DLL")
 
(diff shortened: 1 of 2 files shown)


fujitatomoya (commented):
either @clalancette or @ahcorde can you take a look?

gergondet-woven (approved):
Thanks for the quick response @fujitatomoya I think that would do it for #121 

Could it be cherry-picked to jazzy once it's merged? Thanks!

rcl_logging_noop/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ add_library(${PROJECT_NAME} src/rcl_logging_noop.cpp)
 target_link_libraries(${PROJECT_NAME} PRIVATE
   rcutils::rcutils)
 target_link_libraries(${PROJECT_NAME} PUBLIC
-  rcl_logging_interface::rcl_logging_interface)
+  $<BUILD_INTERFACE:rcl_logging_interface::rcl_logging_interface>)
```
clalancette, line 28: So I'm honestly not sure about all of this, but I don't know that this is correct.  I feel like this will link to the *build* directory, not the install directory, and that doesn't seem correct to me.  Maybe @sloretz can chime in here to give a bit more explanation.

gergondet-woven: If I may, `$<BUILD_INTERFACE:...>` means that the property will be removed when the target is installed via `install(EXPORT)` but available when `export()` is used or when the target is consumed from within the build-tree. ([doc](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html#genex:BUILD_INTERFACE)) So the installed library will link to nothing instead of a library that it doesn't bring in (as part of #102)

sloretz (changes requested):
IIUC because `rcl_logging_spdlog` uses [`target_link_libraries(... PUBLIC)` with a target from rcl_logging_interface](https://github.com/ros2/rcl_logging/blob/4d186fac565caa7dfb2e21b367bec8c074b62835/rcl_logging_spdlog/CMakeLists.txt#L37-L38) it needs to export the dependency on `rcl_logging_interface`. The [package.xml looks ok](https://github.com/ros2/rcl_logging/blob/4d186fac565caa7dfb2e21b367bec8c074b62835/rcl_logging_spdlog/package.xml#L21), but the call to [`ament_export_dependencies(rcl_logging_interface)` line was removed in #102](https://github.com/ros2/rcl_logging/pull/102/files#diff-cfea5d73cea009f5dae4bbd3d73991a6569267fe803d771e9a43992b6ef8ecc5L72)

If adding `ament_export_dependencies(rcl_logging_interface)` does not resolve #121 then there's likely a bug in the CMake scraping done by drake-ros/bazel_ros2_rules.

fujitatomoya:
@gergondet-woven 

I think i was misunderstanding, I am not 100% sure but I think @sloretz 's comment https://github.com/ros2/rcl_logging/pull/122#pullrequestreview-2334345059 is correct. `rcl_logging_spdlog` needs to export the dependency on `rcl_logging_interface`, so putting `ament_export_dependencies(rcl_logging_interface)` back to see if the problem solves?

gergondet-woven:
> so putting ament_export_dependencies(rcl_logging_interface) back to see if the problem solves?

Yes, this would work as well and it's even more correct.

I have also confirmed that I had no scrapping issues in that case since the `rcl_logging_interface::rcl_logging_interface` is correctly imported before the `rcl_logging_spdlog` defines its targets.

fujitatomoya:
@sloretz @gergondet-woven could you review one more time?

gergondet-woven (approved):
LGTM :+1:

ahcorde (approved):

ahcorde:
Pulls: ros2/rcl_logging#122
Gist: https://gist.githubusercontent.com/ahcorde/d01cd595cc9a1be19db5cdb3f823f200/raw/b71a04a9d35d807536727ae4a258d53cd5b6fdbb/ros2.repos
BUILD args:   --packages-above-and-dependencies  rcl_logging_noop rcl_logging_spdlog --packages-above-and-dependencies rcl_logging_noop rcl_logging_spdlog
TEST args:  --packages-above  rcl_logging_noop rcl_logging_spdlog --packages-above rcl_logging_noop rcl_logging_spdlog
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/14635
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/21918/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16184/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1432/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/22639/)

clalancette:
@Mergifyio backport jazzy iron humble

